PR #35, opened 2019-04-23
Issue 706 cached packages

artemdudkin:
**Type:**

PoC version for task "Pages for npm cached packages"
(https://github.com/verdaccio/verdaccio/issues/706)

**Description:**

New page for list of cached packages
verdaccio-pr

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "react-hot-loader": "4.7.1",
     "react-router": "4.3.1",
     "react-router-dom": "4.3.1",
+    "react-virtualized": "9.21.0",
     "resolve-url-loader": "3.0.1",
     "rimraf": "2.6.3",
     "sass-loader": "7.1.0",
--- a/package.json
+++ b/package.json
@@ -120,16 +120,17 @@
     "dev:web": "cross-env BABEL_ENV=ui babel-node tools/dev.server.js",
     "verdaccio:server": "node tools/verdaccio.js",
     "build": "npm run pre:webpack && cross-env BABEL_ENV=ui webpack --config tools/webpack.prod.config.babel.js",
-    "dev": "concurrently --kill-others \"npm run dev:web\" \"npm run verdaccio:server\""
+    "dev": "concurrently --kill-others \"npm run dev:web\" \"npm run verdaccio:server\"",
+    "commitmsg": "commitlint -e $GIT_PARAMS"
   },
   "engines": {
     "node": ">=8",
     "npm": ">=5"
   },
-  "husky": {
-    "hooks": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -119,16 +120,17 @@
...
   },
   "engines": {
     "node": ">=8",
     "npm": ">=5"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "commitlint -e $GIT_PARAMS"
-    }
+  "commitlint": {
```
juanpicado: The topic here is other, I do not disagree, but not the right moment for this. 
```
commitlint": {
    "extends": [
      "@commitlint/config-conventional"
    ]
```

package.json
```diff
@@ -73,6 +73,7 @@
     "react-hot-loader": "4.7.1",
     "react-router": "4.3.1",
     "react-router-dom": "4.3.1",
+    "react-virtualized": "9.21.0",
```
juanpicado: This needs an open discussion, pagination vs infinite scroll. cc: @ayusharma

juanpicado: cc: @verdaccio/maintainers @verdaccio/collaborators

artemdudkin: (1) Right now it gets all data at one request, i.e. without pagination, and it is 18kb response for 3k packages at my local repo (and then was rendered by List component that renders only visible items). It looks ok for me to do it this way now, but if someone will get 10 times more packages with 10 versions for any package then it can be over 1Mb, and this is not good.

(2) If paging arrives then we need "search" endpoint, I guess?

juanpicado: (1) Agree. 
(2) Yes.

I'd go personally with (2), but I'd wait to hear more opinions (hopefully arrive some).

artemdudkin:
Any thoughts?

package.json
DanielRuf: We could stream JSON using Oboe.js or another solution.

juanpicado: Do you have any example that we can take a look such implementation?

DanielRuf: Did not work with oboe.js in the last years.
But npm is also using it as it seems.

http://oboejs.com/examples
https://github.com/jimhigson/oboe.js/network/dependents?dependent_type=PACKAGE
https://github.com/npmtest/node-npmtest-oboe
https://github.com/npm/npmepm/blob/0a37de7aa2cd560836fe6543fd5567e361549e35/lib/dockerps.js#L19-L39

DanielRuf: http://oboejs.com/api

priscilawebdev: I'd love to have pagination in verdaccio ✋

juanpicado:
I close this one because we use already `react-virtualized`

PR #43, opened 2019-05-02
test: fix formatDateDistance test

DanielRuf:
Test formatDateDistance with a date which is about 2 months in the past and another which is exactly 2 months in the past.

**Type:**
unit tests

*

**Description:**

The `formatDateDistance` test was using a date which was exactly 2 months ago so `about 2 months` is not correct but `2 months`. I have added another case where this is `about 2 months` as expected to cover both cases.


DanielRuf:
Seems the commitlint / husky is broken here. It uses my previous message instead of the new one. =(

PR #45, opened 2019-05-02
 feat: browser history is aware of basename

juanpicado:
**Type:** bug

*  There is a related issue? https://github.com/verdaccio/verdaccio/issues/1297
*  Unit or Functional tests are included in the PR No

**Description:**

The UI receives from the application the base name to be used for the History API. This is required otherwise the basename is `/` by default and makes the app non-compatible with reverse proxy using directories.

bug: https://github.com/verdaccio/verdaccio/issues/1297
reproduce error: https://github.com/verdaccio/docker-examples/tree/master/reverse_proxy/nginx/relative_path


tools/webpack.prod.config.babel.js
```diff
@@ -45,7 +45,9 @@ const prodConf = {
     }),
     new HTMLWebpackPlugin({
       title: 'ToReplaceByTitle',
+      __UI_OPTIONS: 'ToReplaceByVerdaccioUI',
```
juanpicado: `__UI_OPTIONS` will replace the options below in future tickets. The current solution does not scale, so, I plan to pass a JS object instead.

ayusharma (approved):

PR #54, opened 2019-05-10
fix: search suggestion container height

ayusharma:
**Type:** bug

Screenshot 2019-05-03 at 23 11 13

**Description:**
- Fixes search container height to 500px.


juanpicado (approved):
👍 tested and works.

src/webui/components/AutoComplete/styles.js
```diff
@@ -50,3 +51,10 @@ export const InputField = ({ color, ...others }: IInputField) => (
     }}
   />
 );
+
+export const SuggestionContainer = styled(Paper)`
+  && {
+    max-height: 500px;
+    overflow-y: scroll;
```
DanielRuf: Should be probably `auto` as otherwise you always have a scrollbar in some browsers.

PR #57, opened 2019-05-12
fix: use auto for overflow-y

DanielRuf:
**Type:**
fix

*  https://github.com/verdaccio/ui/pull/54#discussion_r283093487
* An `overflow` setting of `auto` will always show the scrollbar in many browsers (for example Chrome / Chromium, Firefox and others). See https://codepen.io/DanielRuf/pen/LoZXjY

**Description:**

This sets the correct value for `overflow-y`.


juanpicado:
Thanks @DanielRuf  !

PR #64, opened 2019-05-24
Enable the Node.js 12 actions

DanielRuf:
**Type:**

ci / build

**Description:**

Enables the Node.js 12 actions as we are now at Node.js 12.3.0 on the stable channel.


.github/main.workflow
```diff
@@ -6,7 +6,7 @@ workflow "build and test on branch" {
     "branch.lint.node.10",
     "branch.test.node.10",
     "branch.test.node.8",
-    # "branch.test.node.12"
+    "branch.test.node.12"
```
juanpicado: I think @ayusharma detected an issue with Node12 but I don't remember what exactly.

DanielRuf: I will check this in the next week if there are any issues.

juanpicado:
@DanielRuf I removed testing with GitHub Actions, it does not work fine, hard to debug. Sorry I need to close this, but we can try on CircleCi where Node12 is missing.

PR #73, opened 2019-06-09
fix: fixed failed to load all packages after login

### LINKED ISSUE #72: Failed to load all packages after login
**Describe the bug**
Non-public packages not included in the package-list after login.

**To Reproduce**
1. Load the UI in a browser
2. You can see the public packages available to all
3. Login
4. Package list still the same, not included non-public packages I can access

**Expected behavior**
A package list containing all packages I can access.

> I have a PR ready solving this issue

coolsp:
**fix:**

Non-public packages not included in the package-list after login.

**Description:**

A login forces react to reload the list of available packages (by changing the state).
However the state (which triggers react to refresh) is set before storing the login token resulting in fetching (XHR GET) only the public packages as the token is not yet available in the GET request.

Fixed by setting the storage items ('username' and 'token') before this.setLoggedUser().

Resolves #72

### DIFF
--- a/src/webui/app.js
+++ b/src/webui/app.js
@@ -120,9 +120,9 @@ export default class App extends Component {
     const { username, token, error } = await makeLogin(usernameValue, passwordValue);
 
     if (username && token) {
-      this.setLoggedUser(username, token);
       storage.setItem('username', username);
       storage.setItem('token', token);
+      this.setLoggedUser(username, token);
     }
 
     if (error) {


juanpicado:
Thanks @coolsp , I really liked the detailed description in the commit, awesome 👍

coolsp:
Thanks @DanielRuf @juanpicado for this quick turnaround, great work 🥇 !

juanpicado:
shipped `verdaccio@4.0.2`

Issue #74, opened 2019-06-14
QA: current linting warnings


DanielRuf:
**Describe the bug**
Currently we have several linting warnings and I want to resolve them.
Which of the following warnings are relevant (the consolelog ones probably not at the moment)?

**To Reproduce**
1. `yarn run dev:web`

**Expected behavior**
No linting warnings.

```
 WARNING  Compiled with 18 warnings                                     21:44:51

Module Warning (from ./node_modules/eslint-loader/index.js):

C:\Users\Daniel\GITHUB\ui\src\webui\app.js
  91:7  warning  Unexpected console statement  no-console

✖ 1 problem (0 errors, 1 warning)


Module Warning (from ./node_modules/eslint-loader/index.js):

C:\Users\Daniel\GITHUB\ui\src\webui\components\AutoComplete\index.js
   20:5   warning  spread operator is not allowed jsx elements  verdaccio/jsx-sp
read
   45:13  warning  style attribute is not allowed jsx elements  verdaccio/jsx-no
-style
   49:13  warning  style attribute is not allowed jsx elements  verdaccio/jsx-no
-style
  112:7   warning  spread operator is not allowed jsx elements  verdaccio/jsx-sp
read
  123:7   warning  spread operator is not allowed jsx elements  verdaccio/jsx-sp
read

✖ 5 problems (0 errors, 5 warnings)


Module Warning (from ./node_modules/eslint-loader/index.js):

C:\Users\Daniel\GITHUB\ui\src\webui\components\AutoComplete\styles.js
  23:3  warning  spread operator is not allowed jsx elements  verdaccio/jsx-spre
ad

✖ 1 problem (0 errors, 1 warning)


Module Warning (from ./node_modules/eslint-loader/index.js):

C:\Users\Daniel\GITHUB\ui\src\webui\components\Header\index.js
   33:3   warning  handleLoggedInMenu should be placed after render  react/sort-
comp
  115:9   warning  style attribute is not allowed jsx elements       verdaccio/j
sx-no-style
  115:15  warning  Prop `style` is forbidden on Components           react/forbi
d-component-props

✖ 3 problems (0 errors, 3 warnings)


Module Warning (from ./node_modules/eslint-loader/index.js):

C:\Users\Daniel\GITHUB\ui\src\webui\components\Icon\index.js
  50:5  warning  spread operator is not allowed jsx elements  verdaccio/jsx-spre
ad
[...]

PR #77, opened 2019-06-23
refactor: 74 linting warnings

griffithtp:
**Type:** Refactor with typescript fixes

*  There is a related issue - #74 

**Description:**
As described on #74, linting throws type errors that don't match with defined types when moving across to typescript.
One of them is the `spread operator is not allowed jsx elements` from verdaccio/jsx-spread and can be quite a useful feature to use.

~Currently put it down to 159 warnings. Still work-in-progress.~
Managed to bring it down to 32 wranings and remaining mainly includes:
* [x] @typescript-eslint/no-explicit-any - from src/utils/file-size.ts
* [ ] verdaccio/jsx-spread - should this be a _hard_ rule?
* [ ] verdaccio/jsx-no-style & react/forbid-component-props for style props
* [ ] jest/no-disabled-tests
* [ ] `src/components/Dependencies/Dependencies.tsx` _withRouter()_

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -19,7 +19,7 @@ export const AppContext = React.createContext<null>(null);
 export const AppContextProvider = AppContext.Provider;
 export const AppContextConsumer = AppContext.Consumer;
 
-export default class App extends Component<any, any> {
+export default class App extends Component {
   public state = {
     error: {},
     // @ts-ignore
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -20,10 +20,12 @@ export const AppContext = React.createContext<{}>({});
 export const AppContextProvider = AppContext.Provider;
 export const AppContextConsumer = AppContext.Consumer;
 
-interface AppStateInterface {
+export interface AppStateInterface {
   error?: FormError;
   logoUrl: string;
-  user: {};
+  user: {
+    username?: string;
+  };
   scope: string;
   showLoginModal: boolean;
   isUserLoggedIn: boolean;
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -14,14 +14,24 @@ import '../styles/typeface-roboto.scss';
 import '../styles/main.scss';
 import 'normalize.css';
 import Footer from '../components/Footer';
+import { FormError } from 'src/components/Login/Login';
 
 export const AppContext = React.createContext<{}>({});
 export const AppContextProvider = AppContext.Provider;
 export const AppContextConsumer = AppContext.Consumer;
 
-export default class App extends Component {
-  public state = {
-    error: {},
+interface AppStateInterface {
+  error?: FormError;
+  logoUrl: string;
+  user: {};
+  scope: string;
+  showLoginModal: boolean;
+  isUserLoggedIn: boolean;
+  packages: [];
+  isLoading: boolean;
+}
+export default class App extends Component<{}, AppStateInterface> {
+  public state: AppStateInterface = {
     // @ts-ignore
     logoUrl: window.VERDACCIO_LOGO,
     user: {},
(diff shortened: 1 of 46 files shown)


juanpicado:
@griffithtp awesome! please do not forget to follow https://github.com/verdaccio/contributing/blob/master/CONTRIBUTING.md#commit-guidelines I would not want to `git squash` all your commits.

juanpicado:
⚠️ ⚠️ ⚠️  I'd suggest, **one commit for each rule**, this will make much more easy to track any issue in the future.

DanielRuf:
@griffithtp awesome, your first contribution to Verdaccio =)

@juanpicado squashing is not a problem. I guess you mean rebase + reword.

@griffithtp can you try `git rebase -i HEAD~3` and `reword` the last 3 commits and `git push -f` then or do `git reset HEAD~3` and redo the commits by adding the needed files (just a few) and creating new commits and then run `git push -f`?

Let us know if this is possible or if this is too complicated.

griffithtp:
> ⚠️ ⚠️ ⚠️ I'd suggest, **one commit for each rule**, this will make much more easy to track any issue in the future.

@juanpicado, I can do that if you prefer. I just started from the list of issues and was work the way up the list. If you are going to squash I thought it wouldn't make much difference.
I will do that going forward though, thanks :)

Also, I have a few questions about some rules but can discuss or think through when the number of warnings are much reduced though?
1) Prop `style` is forbidden on Components `react/forbid-component-props` - sometimes it is useful to have simple inline styles?
2) spread operator is not allowed jsx elements  `verdaccio/jsx-spread` - spread props is quite useful, imo

juanpicado:
> One of them is the spread operator is not allowed jsx elements from verdaccio/jsx-spread and can be quite a useful feature to use.

Spread props over component is useful, but hide a lot of context, aside of React also do not recommend use them so much, for that reason I block them using in this project.

https://reactjs.org/docs/jsx-in-depth.html#spread-attributes

It does not mean are 100% forbidden, perhaps there is some corner case, but this rule can be easily disable with comments.

src/App/App.tsx
```diff
@@ -112,7 +124,7 @@ export default class App extends Component<any, any> {
     this.setState(prevState => ({
       // @ts-ignore
       showLoginModal: !prevState.showLoginModal,
-      error: {},
```
juanpicado: Do you remember which rule was yelling here? I do not remember perfectly whether the literal object is required for something else.

juanpicado:
Awesome progress by far !

src/App/App.tsx
griffithtp: My bad error should be removed here instead of setting to undefined since `error` should be optional?
error is changed to an optional type because Typescript throws error when using spread params and it makes more sense if no errors to not set it as an empty object which would change its signature

griffithtp:
> > One of them is the spread operator is not allowed jsx elements from verdaccio/jsx-spread and can be quite a useful feature to use.
> 
> Spread props over component is useful, but hide a lot of context, aside of React also do not recommend use them so much, for that reason I block them using in this project.
> 
> https://reactjs.org/docs/jsx-in-depth.html#spread-attributes
> 
> It does not mean are 100% forbidden, perhaps there is some corner case, but this rule can be easily disable with comments.

They are almost all over the place actually
src/utils/asyncComponent.tsx
src/router.tsx
src/icons/GitHub.tsx
src/components/TextField/TextField.tsx
the list goes on :)

juanpicado:
That's right, for such reason I introduced the rule. We can leave that one for a future PR if you will, I can also help with that one.

griffithtp:
> That's right, for such reason I introduced the rule. We can leave that one for a future PR if you will, I can also help with that one. You decide :-)

good idea :-)
this PR is becoming too big. Should it be merged and create separate PR for the remaining tasks?

juanpicado:
Agree !! merging now, :-) I added you to the organization, it seems the CircleCI is not triggered for forks for some reason, I think you are allowed now to create branches in this project. Thanks for the great effort.

griffithtp:
Thank you that's nice of you :)

PR #80, opened 2019-06-30
Refactor/typescript warning jsx no style

griffithtp:
**Type:** Refactor typescript warnings

*  There is a related issue - #74 

**Description:**
In this PR we remove inline-style css following `verdaccio/jsx-no-style` rule;
We use emotion css method however emotion will wrap in class names when generating snapshot.
This should be no issue when replacing node-sass with emotion.

**Resolves:** #74


juanpicado:
LGTM @griffithtp 👍 thanks so much for continue cleaning this up.

Issue #78, opened 2019-06-24
Replace node-sass by emotion


juanpicado:
**Is your feature request related to a problem? Please describe.**

We want to remove `node-sass` from our code base, actually, we have 2 CSS libraries living in our code base. 

**Describe the solution you'd like**
I want to migrate all to `emotion` and reduce dependency from `sass`.

**Describe alternatives you've considered**

We are considering `material-ui` styles as an alternative to keep all in the same world, but that will be under discussion.

**Additional context**

List of current files using `scss` files.

<img width="1016" alt="Screen Shot 2019-06-24 at 8 59 59 PM" src="https://user-images.githubusercontent.com/558752/60044650-1ebcf580-96c3-11e9-8296-406e9dc4ba0b.png">

griffithtp:
If there is no takers, I would be keen to give it a try?

juanpicado:
All yours @griffithtp , I think nobody is actively working on this as far I know.

griffithtp:
@juanpicado we might also need to upgrade to emotion v10 here and it has some breaking changes I am looking at to resolve.
Also, snapshot testing doesn't seem to work using the jest-emotion serializer for components that already has a render method, which is a shame; will look into further config settings.
If anyone has experienced a similar issue, do let us know?

juanpicado:
I would wait for that upgrade, we have been considering use https://material-ui.com/es/styles/basics/ instead which is built-in in material-ui. @priscilawebdev suggested to  use it, back then was in alpha so we stopped, but now it is stable.

griffithtp:
@juanpicado can this be closed and resolved now? 😃

juanpicado:
Thanks @griffithtp :-) you are right !

PR #85, opened 2019-07-05
refactor: 78 replace node sass with emotion

griffithtp:
**Type:** Refactor

*  There is a related issue - #78 

**Description:**
There is 2 CSS libraries living in our code base and this PR is removing node-sass library in favor of emotion v9.
Note: there are some thoughts to move to Material UI Styles
https://github.com/verdaccio/ui/issues/78#issuecomment-508325094

**Resolves** #78


juanpicado:
Awesome @griffithtp ! If there is nothing else remaining I guess we can remove `node-sass`, `sass-loader`, `stylelint-scss`, `stylelint-config-recommended-scss` from `package.json` and from the lock file.

src/App/styles.ts
```diff
@@ -0,0 +1,17 @@
+import { css } from 'emotion';
+import colors from '../utils/styles/colors';
+
+export const alertError = css`
+  background-color: ${colors.red} !important;
+  min-width: inherit !important;
+`;
+
+export const alertErrorMsg = css`
+  display: flex;
```
priscilawebdev: I have seen that in some places you are using template literals and in other places an object ... I think we should decide which one to use and follow a standard

griffithtp: Glad you raise this point :)
I don't have a strong opinion on this.
Template literals feels more natural to me because it is closer to scss with simple styles. Object may be more optimal? when parsing a function.
Thoughts?

priscilawebdev: Hi.Thank you for asking..I would rather use objects ... since we can benefit from autocompletion and since we plan to migrate to material-ui/styles. 😉

juanpicado: Objects I think would be the way (not 100% sure), but for this PR template literals are more than enough, we use them all over the place.

juanpicado:
We have small issue, the fonts are not properly applied. 
Screen Shot 2019-07-07 at 12 39 27 PM

src/App/styles.ts
griffithtp: I have updated the simple ones below.
Perhaps a slow transition as more and more Material UI components gets added?

juanpicado:
The issue might be we were applying some styles to body check https://github.com/verdaccio/ui/pull/85/files#diff-c4de5453df6c069e26c222e36ee93651

juanpicado (changes requested):
Let me know wether I can assist you in something else

src/App/styles.ts
griffithtp: fyi, have reverted again. I can't find a way to remove the compilation error it throws for some odd reasons if all exports are css objects
```
[0] src/components/Login/styles.ts
[0]  1:1  ✖  Unexpected empty source   no-empty-source
```

juanpicado: @griffithtp switch `no-empty-source` to warning for now in the `.eslintrc` folder, we will investigate that later on.

griffithtp:
@priscilawebdev @juanpicado, is this redundant since moving to Material-UI?
sad to have wasted time :'(

juanpicado:
It is fine, emotion is the main library now. Material styles is a good option but that requires a lot of research before move there. Your PR is highly valuable. 🤘🏼

juanpicado:
I also detected a difference with the layout, this PR vs the URL below. 
Kapture 2019-07-07 at 21 18 54

https://registry.verdaccio.org/-/web/detail/webpack

griffithtp:
thanks @juanpicado, should be fixed now

juanpicado:
thank you @griffithtp I'll review it now

juanpicado:
@griffithtp please read this
https://github.com/verdaccio/ui/pull/90#issuecomment-509116844
I need to hold these unfortunately.

griffithtp:
> @griffithtp please read this
> [#90 (comment)](https://github.com/verdaccio/ui/pull/90#issuecomment-509116844)
> I need to hold these unfortunately until I verify the source of the issues.

Please let know if there is anything else to do or test here? 😃

juanpicado:
I will @griffithtp , I just want to be sure I don't report to you fix bugs you did not created. :-) it will be unfair.

juanpicado (approved):
:-) 🔍 juan seal of approval 💯 . Nice work, bye bye Sass

src/App/App.tsx
```diff
@@ -25,6 +25,7 @@ export interface AppStateInterface {
   logoUrl: string;
   user: {
     username?: string;
+    token?: string;
```
juanpicado: Opps @griffithtp could we remove this?

juanpicado (changes requested):
I just saw this, sorry.

src/App/App.tsx
priscilawebdev: After that I can also approve...

src/App/styles.ts
```diff
@@ -0,0 +1,17 @@
+import { css } from 'emotion';
+import colors from '../utils/styles/colors';
+
+export const alertError = css`
+  background-color: ${colors.red} !important;
```
priscilawebdev: I don't use "!important;" anymore..I think when you need to use it...something is wrong..

griffithtp: Completely agree with you on this. 
This was a copy from the scss files I didn't want to mix up with actually refactoring styles. I think this most likely will be canned with Material-UI?

sergiohgz: In case we need this, we can surround in `&&` block to repeat the css class, making it important without explicitly writing the taboo word (!important)
```css
&& {
  background-color: ${colors.red};
}
```

src/App/App.tsx
```diff
@@ -138,7 +132,7 @@ export default class App extends Component<{}, AppStateInterface> {
     if (username && token) {
       storage.setItem('username', username);
       storage.setItem('token', token);
-      this.setLoggedUser(username, token);
+      this.setLoggedUser(username);
```
griffithtp: Also removed here since no longer set in AppState.
I'm thinking in this case it might be good to use the `setLoggedUser()` as the single function to handle username, token, etc. therefore move the `storage.setItem()`?
Perhaps a different PR without deviating too much from this one 😄

src/App/styles.ts
sergiohgz: Ouch, I saw `!important` in more files, so we can keep it for a while and then remove, it isn't in the scope of the PR (great work, I'm reviewing :wink:)

src/App/App.tsx
juanpicado: For now it is ok, agree with you, a new PR but I will read that logic again later on.

sergiohgz (approved):
:1st_place_medal: :clap::clap::clap::clap:

src/App/styles.ts
juanpicado: Agree, I dislike important, which sounds funny. But, it was there already so, cool for now. Legacy stuff.

juanpicado (approved):
🤘🏼

juanpicado:
Sergio, for some reason workflow does not get reported in PR, this is odd, sometimes does and sometimes does not. :(

sergiohgz:
> Sergio, for some reason workflow does not get reported in PR, this is odd, sometimes does and sometimes does not. :(

@juanpicado It seems that external PR are not running the workflow in CircleCI, I will investigate how we did in other repos

sergiohgz:
I've just enabled forked PR execution on CircleCI, it was disabled. @griffithtp, can you push an empty commit to try (if you want, if not, I'll fork and create a fake PR to test this behaviour :wink:)

circleci_status_after_enable

cc: @juanpicado

sergiohgz:
Wow, it works, kudos @juanpicado for the advice and @griffithtp for the test :blush:

juanpicado:
@sergiohgz I think we might have the same issue in other repos.

@griffithtp awesome, thanks so much !!

PR #86, opened 2019-07-06
fix: 83 support deprecated license object properties

### LINKED ISSUE #83: Crashing while trying to view readme
**Describe the bug**
Using v4.04 -
Just clicking on a link to view the readme throws the following error in the browser (both brave and firefox):
```
vendors.b6891b53f2830d5ac3cb.js:46 Invariant Violation: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Btype%2C%20url%7D&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.   
```

Crashing Browser side Code:
```
   function fo(e, t) {
        "textarea" !== e.type && a("31", "[object Object]" === Object.prototype.toString.call(t) ? "object with keys {" + Object.keys(t).join(", ") + "}" : t, "")
    }
```
The e.type was element of type "span".  
The t = was  the license section from the package.json: 
```
"license": {
    "type": "scoped",
    "url": "https://git.master.technology/scoped/somename-month-calendar/blob/master/LICENSE"
  },
```

**To Reproduce**
1. I logged in via npm 
2. I uploaded a single @scoped package
2. Uplink is disabled in config
3. I navigated to the main page (that lists plugins)
4. I clicked the package name to see the readme  (i.e. link is `/-/web/detail/@scoped/somename-month-calendar`
5. See error

**Expected behavior**
I assume I would see the package readme.

**Configuration File (cat ~/.config/verdaccio/config.yaml)** 

I removed the other package sections; as I only need this one.   (No uplink active), no other packages.  The rest of the config except for the listen is stock.

```
packages:
  '@scoped/*':
    # scoped packages
    access: $all
    publish: $authenticated
```

**other info**
All http requests were status 200 or 304 on both browser and console on server.

griffithtp:
**Type:** bugfix

*  There is a related issue? *Yes*
*  Unit or Functional tests are included in the PR. *Yes*

**Description:**
* Add support for license object properties
* Add unit test for <Dist/> Component
* Refactor DistChip renderer

Resolves #83


types/packageMeta.ts
```diff
@@ -5,7 +5,12 @@ export interface PackageMetaInterface {
       fileCount: number;
       unpackedSize: number;
     };
-    license: string;
+    license?: Partial<LicenseInterface> | string | null;
```
priscilawebdev: Hmm ... do we really need to have "null" here? I think if the license is undefined it is enough 🙃

griffithtp: I agree and I thought of that at first, however there are subtle differences between undefined and null :)
Also, the reason here is because *formatLicense()* may actually return null, which is better than empty string
`https://github.com/verdaccio/ui/blob/4.x-master/src/utils/package.ts#L20-L30`

priscilawebdev: Hi! Yes there are differences between null and undefinied :-) in this case I would update _formatLicense()_ to return undefinied...you can write as below:

`export function formatLicense(license: any): string | undefined {
  if (isString(license)) {
    return license;
  }

  if (license && isObject(license) && license.type) {
    return license.type;
  }

  return ;
}`

I have also seen that the license is been used here - only:
image

So undefined will be better

juanpicado (approved):
👏 💯

juanpicado:
Thanks @griffithtp for fixing this :-)

griffithtp:
> Thanks @griffithtp for fixing this :-)

all good @juanpicado 
if there is any tasks you can assign and I will be happy to take a look.
roadmap might seem a bit unclear at the moment? 😄

PR #90, opened 2019-07-07
Chore: Migrate Material UI Dependencies to version 4.X.X

priscilawebdev:
**Type:** Chore

The following has been addressed in the PR:  closes #85 

*  There is a related issue?
No

No..it was not necessary

**Description:**

As part of our process, we try to keep our libraries up to date. In this PR the updated libraries were:

- @material-ui/core
- @material-ui/icons
- @types/enzyme

As the version 4 of Material UI has breaking changes, we followed this guideline: https://material-ui.com/guides/migration-v3/

As we are still not using theme, the breaking changes didn't affect us much, with the exception of forwardRef. It was necessary to include ForwardRef in the following components:

- IconButton
- Avatar
- Link

**Why did I move some components to the "primitive" folder:** 

I found necessary to move the above components to the "primitive" folder, because I was getting confused. Some  of the components are Material UI - Only and some are verdaccio's patterns.

I also think that once we have a theme, we need to limit some of the passed props, so that's one more reason to keep the primitive material components in a separate folder.

**Why did I add the rule  "react/display-name":** 

Well, I think there is a problem (I will investigate more), but it is not necessary to always add a component name, if it has already one.


src/components/DetailSidebar/DetailSidebar.tsx
```diff
@@ -5,7 +5,7 @@ import CardContent from '@material-ui/core/CardContent';
 import List from '@material-ui/core/List';
 
 import ActionBar from '../ActionBar/ActionBar';
-import Author from '../Author';
+// import Authors from '../Author';
```
sergiohgz: The author component is not rendered any more in this sidebar as it’s commented. Is this correct?

src/components/Header/Header.tsx
```diff
@@ -174,9 +185,11 @@ class Header extends Component<Props, State> {
       case 'help':
         content = (
           // @ts-ignore
-          <IconButton blank={true} color={'inherit'} component={ExternalLink} to={'https://verdaccio.org/docs/en/installation'}>
-            <Help />
-          </IconButton>
+          <ExternalLink blank={true} to="https://verdaccio.org/docs/en/installation">
```
sergiohgz: Is this correct? We should use ForwardRef to make button behaviour like a anchor link, not wrap the button with a link (accessibility here is a good point)

src/components/DetailSidebar/DetailSidebar.tsx
priscilawebdev: Good catch! that is the reason I still like to have Snapshots...thank you :-)

src/components/Author/Author.test.tsx
```diff
@@ -27,8 +27,7 @@ describe('<Author /> component', () => {
     }));
 
     const Author = require('./Author').default;
-    const wrapper = shallow(<Author />);
-    expect(wrapper.html()).toMatchSnapshot();
+    expect(<Author />).toMatchSnapshot();
```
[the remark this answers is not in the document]
priscilawebdev: @sergiohgz that is true...reason Snapshots are still good to have..thank you :-) Well I am getting an error when I shallow this component. it doesn't accept react hooks...do you have some idea of how to fix this issue? thanks 👍 :-)

src/components/Header/Header.tsx
priscilawebdev: We need a ForwardRef because the component is being used as a child of ToolTip...
The IconButton component hasn't the same props anymore...as before...so I also removed invalid props

src/components/Author/Author.test.tsx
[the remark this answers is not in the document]
sergiohgz: I’ve just read that Enzyme has support for testing hooks (but don’t complete and I don’t know if they are released as stable) and some patterns like wrapping hooks in a component to test how they work. Do that help you? I don’t have a computer near to investigate more 😢

src/components/Author/__snapshots__/Author.test.tsx.snap
```diff
@@ -1,5 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`<Author /> component should render the component in default state 1`] = `"<ul class=\\"MuiList-root-1 MuiList-padding-2 MuiList-subheader-4\\"><h3 class=\\"MuiTypography-root-5 MuiTypography-subheading-12 css-hyrz44 e1xuehjw0\\">Author</h3><li class=\\"MuiListItem-root-41 MuiListItem-default-44 MuiListItem-gutters-49 css-z8a2h0 e1xuehjw1\\"><a href=\\"mailto:verdaccio.user@verdaccio.org?subject=verdaccio@4.0.0\\" target=\\"_top\\"><div class=\\"MuiAvatar-root-53\\"><img alt=\\"verdaccio user\\" src=\\"https://www.gravatar.com/avatar/000000\\" class=\\"MuiAvatar-img-55\\"/></div></a><div class=\\"MuiListItemText-root-56\\"><span class=\\"MuiTypography-root-5 MuiTypography-subheading-12 MuiListItemText-primary-59\\">verdaccio user</span></div></li></ul>"`;
+exports[`<Author /> component should render the component in default state 1`] = `<Authors />`;
```
ayusharma: 🚨 You are removing snapshots. Any reason ??

juanpicado:
We should stick with the main topic of the PR, migrate the library. I haven’t look yet the migration guide but (as I already discuss with Pris) eslint and other things must be addressed in a different PR. We need to review why the coverage decreased 2% and please we should  not modify any unit test if is not required or topic related (unless is documented the reason). 

Let’s try keep things small so is easy to review. 🙏 (on my mobile now) I will take a look the changes later on.

.eslintrc
```diff
@@ -85,7 +85,7 @@
     "react/jsx-no-bind": ["error"],
     "react/jsx-no-comment-textnodes": ["error"],
     "react/jsx-no-duplicate-props": ["error"],
-    "react/jsx-no-literals": ["error"],
```
ayusharma: Do we need to change eslint rules to update new version of material-ui ? I am not clear about this change.

src/components/Author/__snapshots__/Author.test.tsx.snap
priscilawebdev: Have you seen my answer to @sergiohgz feedback? :-)

src/components/Author/Author.test.tsx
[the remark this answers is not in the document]
priscilawebdev: image
 :/

priscilawebdev:
Thank you for your feedback! 👍 I already removed the changes I did in the eslint file (now tests are broken), but I said previously, I was going to open new PR's for the changes.

Regarding the test I could not make shallow or mount from enzyme play well with react hooks...Please see: https://github.com/reactjs/rfcs/issues/71..I appreciate if you have some suggestion...

ps: As I suggested before .. I would rather use react-testing-library..It fits better our needs..and plays well with hooks

https://github.com/testing-library/react-testing-library

.eslintrc
priscilawebdev: I was getting an error, due to the ForwardRef I added in some components ... it was teelling me that my functional components should be named...something like Link.name = 'Link' what I find meaningless.

juanpicado:
I'll hold this until until we fix all bugs, currently the master is blocked since we have visual bugs and I could not add the latest version to the `verdaccio@4.1.0` release. 
https://github.com/verdaccio/verdaccio/commit/44c16107911916b4b7964e7f29471f3138144c95

juanpicado:
I would rather wait for this https://github.com/airbnb/enzyme/issues/2011 

> ps: As I suggested before .. I would rather use react-testing-library..It fits better our needs..and plays well with hooks

Enzyme will support hooks soon or later, I'd prefer put the effort in their shoulders rather in ours.  So, let's wait for it 🍿 .

priscilawebdev:
@juanpicado ok so let's wait 😉

PR #91, opened 2019-07-08
feat: updated typescript to 3.5.2

tmkn:
**Type:**

*  Unit or Functional tests are included in the PR No

**Description:**
This increases the version of TypeScript to 3.5.2
Due to the improved type checking some adaptions were necessary
Tests have been adapted


src/App/App.tsx
```diff
@@ -86,7 +86,7 @@ export default class App extends Component<{}, AppStateInterface> {
       this.handleLogout();
     } else {
       this.setState({
-        user: { username, token },
```
juanpicado: :-) we need the token otherwise you cannot login.

sergiohgz: Why removing the token from the App state?

tmkn: sure, but `token` is not defined in `AppStateInterface`

src/App/App.test.tsx
```diff
@@ -71,7 +71,6 @@ describe('App', () => {
     await handleDoLogin('sam', '1234');
     const result = {
       username: 'sam',
-      token: 'TEST_TOKEN',
```
sergiohgz: Same here, token is necessary, so the test shouldn't be modified

juanpicado:
For this PR I'd rather use the commit message `build: xxx` since it is not a feature user can actually use, it is only visible for us, developers :-)

src/App/App.tsx
tmkn: it gets written to `storage.setItem('token', token);` instead

yarn.lock
```diff
@@ -13059,10 +13059,10 @@ typeface-roboto@0.0.54:
   resolved "https://registry.npmjs.org/typeface-roboto/-/typeface-roboto-0.0.54.tgz#8f02c9a18d1cfa7f49381a6ff0d21ff061f38ad2"
   integrity sha512-sOFA1FXgP0gOgBYlS6irwq6hHYA370KE3dPlgYEJHL3PJd5X8gQE0RmL79ONif6fL5JZuGDj+rtOrFeOqz5IZQ==
 
-typescript@3.4.5:
-  version "3.4.5"
-  resolved "https://registry.verdaccio.org/typescript/-/typescript-3.4.5.tgz#2d2618d10bb566572b8d7aad5180d84257d70a99"
-  integrity sha512-YycBxUb49UUhdNMU5aJ7z5Ej2XGmaIBL0x34vZ82fn3hGvD+bgrMrVDpatgz2f7YxUMJxMkbWxJZeAvDxVe7Vw==
+typescript@3.5.2:
+  version "3.5.2"
+  resolved "https://registry.npmjs.org/typescript/-/typescript-3.5.2.tgz#a09e1dc69bc9551cadf17dba10ee42cf55e5d56c"
```
sergiohgz: Can you point to https://registry.verdaccio.org when installing to cache in the registry?

juanpicado:
@priscilawebdev do you see any conflict with Material-UI using the latest TS version?

src/App/App.tsx
juanpicado: I think might be a mistake. Let's wait @tmkn reply.

tmkn: Because the `token` itself is not defined in the `AppStateInterface` (I didn't change the interface)
```ts
export interface AppStateInterface {
  error?: FormError;
  logoUrl: string;
  user: {
    username?: string;
  };
  scope: string;
  showLoginModal: boolean;
  isUserLoggedIn: boolean;
  packages: [];
  isLoading: boolean;
}
```
only `username` is

yarn.lock
juanpicado: @sergiohgz unfortunately that hard to enforce, it depends rather of us to use `.nmprc`. But I'm ok if the change its back.

src/App/App.tsx
juanpicado: Good point, I think should be defined there. You just found a bug :-)

yarn.lock
juanpicado: Some context for @tmkn, we use https://registry.verdaccio.org/ as proxy registry to avoid any 500/404 npmjs affect us. But we are not enforcing such thing via config files yet.

tmkn: ok, if you can tell me how to fix this/avoid this in the future?

juanpicado: @tmkn no worries, just leave it like this. I'll take care of this in other PR. It should be transparent for you.

src/App/App.tsx
sergiohgz: Yeah, good catch @tmkn 😄

yarn.lock
sergiohgz: A simple way to make this is yarn install —registry `https://registry.verdaccio.org`

src/App/App.tsx
tmkn: As far as I can see, the `token` is only ever written to the App state, never read from, it's written and read from, from the `storage` object. instead. (`storage.getItem('token')` etc)

Also only 1 test failed(which was just testing the presence of the token), so I assumed that ts is right and the token doesn't belong into the App state and removed the assignments.

But I can add them back of course

griffithtp: Good find.
I also found it missing and added https://github.com/verdaccio/ui/pull/85/files#diff-2d39dbde05630bdfe781f3a465e7857bR28

yarn.lock
griffithtp: adding `.npmrc` would fix it?

src/App/App.tsx
juanpicado: You are completely right @tmkn 🚀 . It is useless. We persist with local storage. Go ahead then.

juanpicado: Good catch @griffithtp !

PR #92, opened 2019-07-08
chore: adds unit tests for <Engines /> component

ayusharma:
**Type:** unit test

*  There is a related issue?  **No**
*  Unit or Functional tests are included in the PR  **Yes**

**Description:**
- Adds unit tests for **Engines** component


src/components/Engines/Engines.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      },
+    };
+
+    jest.doMock('../../pages/version/Version', () => ({
+      DetailContextConsumer: component => {
+        return component.children({ packageMeta });
+      },
+    }));
+
+    const Engines = require('./Engines').default;
```
sergiohgz: Is this necessary or we can import globally with ESM import syntax?

src/components/Engines/Engines.test.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+jest.mock('./img/node.png', () => '');
```
juanpicado: We should not mock this due to we uses `identity-obj-proxy` to do it out of the box, but If I remove the mock the test fails, thus, the configuration seems does not work fine. I'll address this in a new ticket.
https://github.com/keyz/identity-obj-proxy
See example here.
https://github.com/verdaccio/ui/blob/4.x-master/jest.config.js#L16

juanpicado: https://github.com/verdaccio/ui/issues/93

src/components/Engines/Engines.test.tsx
ayusharma: We need this for `jest.doMock`. This is locally scoped mock. Thanks for pointing out 😄

src/components/Engines/Engines.test.tsx
ayusharma: Yes. You are right. I'll investigate it. Nice catch 👍

src/components/Engines/Engines.test.tsx
juanpicado: @sergiohgz it is, check here https://jestjs.io/docs/en/jest-object.html#jestdomockmodulename-factory-options

It took me a while to understand this, Jest mock hoist imports, but when you want to use `jest.doMock` things are different, you need to reimport with `require` to get the mocked import properly, otherwise you get the original one from the imports hoisted above. So this configuration is correct.

juanpicado: https://medium.com/@elderfo/collocating-integration-and-unit-tests-using-jest-domock-34e76cf14f6e

sergiohgz: All right, I understand know. But, if the png files are mocked with `identity-obj-proxy`, we can move to ESM import syntax, am i right?

sergiohgz: All right, I understand, as we want to mock some underlying components with `doMock`, we need this require, to load the component after mocking. Good knowledge to go bed in a while :smiley:

ayusharma: Yes. I really like `doMock` API. It is super handy most of the time.

PR #47, opened 2019-05-03
Refactor: migrating Flow to Typescript

priscilawebdev:
This PR convert the code base to Typescript, the changes are the following:

- [x] migrate code base to Typescript (3.4.x)
- [x] enable `eslint` and `@typescript-eslint/eslint-plugin` (warnings still need to be addressed in future pull request
- [x] update relevant dependencies for this PR (linting, etc)
- [x] enable `bundlezise` (it was disabled for some reason)

PR #94, opened 2019-07-09
fix: incorrect logos styles

### LINKED ISSUE #89: The logos has incorrect style
**Describe the bug**

We have regression issues due to https://github.com/verdaccio/ui/pull/80

Verdaccio_-_webpack

The issue seems we forgot to remove some quotes https://github.com/verdaccio/ui/pull/80/files#diff-2780c4d0882ad10abae71aa33af34c7bR147

It affects loading logo and header logo.

Issues

Screen Shot 2019-07-07 at 1 02 02 PM

Screen Shot 2019-07-07 at 1 01 14 PM

sergiohgz:
**Type: fix**

*  There is a related issue? #89 
*  Unit or Functional tests are included in the PR: Updated snapshots related

**Description:** This fixes the logo styles following next:
1. Remove unnecessary quotes that surrounds value in Emotion css templated string. This happens only when using templated string, with object syntax quotes are mandatory to transform value in a string.
2. Revert a change made in #47, where `md` prop where removed from Logo component. After revert and set a bigger size when `md` (the same value before that PR), logo is centered and have correct styles.


priscilawebdev:
@sergiohgz Thank you!!

src/components/Loading/Loading.tsx
```diff
@@ -8,7 +8,7 @@ import { Wrapper, Badge } from './styles';
 const Loading: React.FC = () => (
   <Wrapper>
     <Badge>
-      <Logo />
+      <Logo md={true} />
```
griffithtp: what do you think about a `size` prop taking enums instead which will allow multiple size in future for the Logo component?

juanpicado: I'd rather remove the parameter, I think we will never have `90px` there. We even suggest use `40px` in the documentation.

juanpicado: Aside of that, it works perfect now.

sergiohgz: Yeah, that approach was there before #47, so the first step is revert to that moment and later improve with an enum. I've just updated, take a look again @juanpicado @griffithtp

Issue #76, opened 2019-06-20
Download button is not displayed if the tarball dist URI has localhost as domain


juanpicado:
**Describe the bug**

The download button is hidden if I run `verdaccio` as `localhost`.

The reason is the `validator` library and the current setup, validate 

```
"http://localhost:8080/@verdaccio%2fui-theme/-/ui-theme-0.2.0.tgz"
```
as false.
```
export function isURL(url): boolean {
  return isURLValidator(url || '', {
    protocols: ['http', 'https', 'git+https'],
    require_protocol: true,
  });
}
```

**To Reproduce**

1. Publish any package
2. Go to the detail page
3. The button is not being rendered

**Expected behavior**

In localhost environments, this should be valid.

**Screenshots**

![Screen Shot 2019-06-20 at 5 26 39 PM](https://user-images.githubusercontent.com/558752/59861166-9bc53380-9380-11e9-825c-1db2f1ecf376.png)

**Configuration File (cat ~/.config/verdaccio/config.yaml)** 
Packages protected
```
packages:
  '@*/*':
    access: $authenticated
    publish: $authenticated
    unpublish: $authenticated
    proxy: npmjs

  '**':
    access: $authenticated
    publish: $authenticated
    unpublish: $authenticated
    proxy: npmjs
```

**Additional context**

DanielRuf:
localhost should be a valid URL, see https://codepen.io/DanielRuf/pen/dBNVMQ?editors=1111

DanielRuf:
Not a great solution, see https://github.com/chriso/validator.js/issues/757

juanpicado:
I think you got a point, I was happy with the dependency, something less to maintain, but seems to be another layer of complexity.

griffithtp:
@DanielRuf's proposed solution with require_tld is probably the way to go?
babel polify still means we have one package to maintain and it appears there is an issue with it though this seems outdated?
https://github.com/babel/babel/issues/7300

I found another from Financial-Times
https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/URL
Yet again something to maintain.

Thoughts anyone?

juanpicado:
@griffithtp I'd go with something we do not have to maintain, add TS types or something like that. :-) the idea is has less code as possible.

DanielRuf:
Why not directly integrate the polyfill? I think I have linked some ppolyfill some time ago.

Polyfills are better than such big solutions like `validator`.

juanpicado:
[...]

PR #101, opened 2019-07-13
fix: 76 download button hidden for localhost

griffithtp:
**Type:** bugfix

*  There is a related issue? #76 
*  Unit or Functional tests are included in the PR - YES

**Description:**
As described in the attached issue 76, download tarball files were not showing for localhost as it is not TLD.
Also, in this PR we refactor url.ts and history.ts by removing unused functions and files being called once only.

**Resolves** #76

### DIFF
--- a/src/router.tsx
+++ b/src/router.tsx
@@ -2,11 +2,15 @@
 
 import React, { Component, ReactElement } from 'react';
 import { Router, Route, Switch } from 'react-router-dom';
+import { createBrowserHistory } from 'history';
 import { AppContextConsumer, AppStateInterface } from './App/App';
 
 import { asyncComponent } from './utils/asyncComponent';
-import history from './history';
 import Header from './components/Header';
+
+const history = createBrowserHistory({
+  basename: window.__VERDACCIO_BASENAME_UI_OPTIONS && window.__VERDACCIO_BASENAME_UI_OPTIONS.url_prefix,
+});
 
 const NotFound = asyncComponent(() => import('./components/NotFound'));
 const VersionPackage = asyncComponent(() => import('./pages/version/Version'));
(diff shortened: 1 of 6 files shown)


griffithtp:
Added unit test for isURL() and isEmail() however I am not entirely sure there is a purpose having `getBaseNamePath()` and `getRootPath()` in the file.
Shall we remove them if it's used only once, the other not used?

juanpicado:
The method `getBaseNamePath` is definitely used, but about `getRootPath` I'd like to investigate what happened with the place was this being used before remove it.

griffithtp:
> The method `getBaseNamePath` is definitely used, but about `getRootPath` I'd like to investigate what happened with the place was this being used before remove it.

`getBaseNamePath()` only used in history.ts and unless we can anticipate future use it might simpler without it and consolidate test on history.test.ts. 
```
const history = createBrowserHistory({
  basename: window.__VERDACCIO_BASENAME_UI_OPTIONS && window.__VERDACCIO_BASENAME_UI_OPTIONS.url_prefix,
});
```

Anyway this perhaps is irrelevant in this PR 😄

juanpicado:
ahh, let's do it now, make sense, :-)

src/router.tsx
```diff
@@ -2,12 +2,16 @@
...
 import { Router, Route, Switch } from 'react-router-dom';
+import { createBrowserHistory } from 'history';
 import { AppContextConsumer, AppStateInterface } from './App/App';
 
 import { asyncComponent } from './utils/asyncComponent';
-import history from './history';
 import Header from './components/Header';
 
+const history = createBrowserHistory({
+  basename: window.__VERDACCIO_BASENAME_UI_OPTIONS && window.__VERDACCIO_BASENAME_UI_OPTIONS.url_prefix,
```
DanielRuf: Do we need the trailing comma?

griffithtp: linting rule

juanpicado:
Thanks @griffithtp  🚀

PR #110, opened 2019-07-28
fix: download protected tarballs

### LINKED ISSUE #75: Cannot download tarball after logging in
**Describe the bug**
Access to packages is restricted to authenticated users, which works fine during browsing. However, on clicking "Download tarball" I get an error:
```json
{
  "error": "authorization required to access package my-package"
}
```

**To Reproduce**
1. Login
2. Navigate to a package with access `$authenticated`
3. Click on "Download tarball"

**Expected behavior**
The tarball should be downloaded.

**Configuration File (cat ~/.config/verdaccio/config.yaml)** 
```yaml
packages:
  '**':
    access: $authenticated
    publish: special
    unpublish: special
    # if package is not available locally, proxy requests to 'npmjs' registry
    proxy: npmjs
```

**Additional context**
I checked the network requests and the download does not contain an `Authorization` header.

juanpicado:
**Type:** bug

*  There is a related issue? yes

**Description:**

- fix issue with webpack and the usage of port 8080, not need it anymore for UI (calls are proxied correctly) https://github.com/verdaccio/ui/pull/110/files#diff-c6c55c02902deda0abd6ef54eb21384eR35
- fix issue where tokens are not being send it to the endpoint (🔴 master **is** broken)
- fix issue downloading tarballs which are protected (after login) #75

The solution was not that easy as I assumed, as you might know, download a tarball requires an Authorization 
header with the token, this is not possible via browser GET, thus, I had to fetch a blob and download it from it adding the right headers.

https://verdaccio.nyc3.digitaloceanspaces.com/gif/Kapture%202019-07-28%20at%2014.11.30.gif

Resolves #75


DanielRuf:
> this is not possible via browser GET

Can we not use `e.preventDefault()` + `fetch`?

juanpicado:
> > this is not possible via browser GET
> 
> Can we not use `e.preventDefault()` + `fetch`?

Do you mean?
```
 onClick={(e) => {
   e.preventDefault();
  /* eslint-disable @typescript-eslint/no-non-null-assertion */
  actionItem.handler!(link);
}}
```

DanielRuf:
Not sure what the following is

> actionItem.handler!(link)

I mean native `window.fetch` which can set headers like any AJAX request.

src/utils/api.ts
```diff
@@ -21,7 +21,7 @@ function handleResponseType(response: Response): Promise<[boolean, Blob | string
     }
 
     // unfortunatelly on download files there is no header available
-    if (response.url.match(/tgz/).length > 0) {
+    if (response.url && response.url.match(/.tgz/) !== null) {
```
DanielRuf: Do they end with this substring?

juanpicado: All the time.

DanielRuf: Why not `endsWith(string)`?

juanpicado: That's better 🙃

src/components/ActionBar/ActionBar.tsx
```diff
@@ -54,16 +73,34 @@ class ActionBar extends Component {
...
-        );
+        const actionItem: Action = ACTIONS[value];
+        if (actionItem.handler) {
+          const fab = (
+            <Tooltip key={key} title={actionItem['title']}>
+              <Fab
+                /* eslint-disable react/jsx-no-bind */
+                onClick={() => {
+                  /* eslint-disable @typescript-eslint/no-non-null-assertion */
+                  actionItem.handler!(link);
```
juanpicado: There is a check for `handler` above, I don't know why `eslint` does not understand that, so I used the null assertion to bypass the issue.

sergiohgz (approved):
LGTM

PR #115, opened 2019-08-12
hooks first step

juanpicado:
**Type:** refactor

The candidate here is `src/components/Developers/Developers.tsx`. 
I extracted the Avatar to `src/components/AvatarTooltip/AvatarTooltip.tsx`  🆕 

*  There is a related issue? 
*  Unit or Functional tests are included in the PR yes

**Description:**

First try using hooks.  Some notes:

- By now, only `useState` and `useContext` usage.
- Enzyme support hooks if you `mount` the component.
- `@testing-library/react` might be useful in coordination with `Enzyme`. 
- We can move to hooks only components if update the test to `mount` 
- It might be verbose due material-ui using `mount`.


package.json
```diff
@@ -16,6 +16,7 @@
     "@material-ui/core": "3.9.3",
     "@material-ui/icons": "3.0.2",
     "@octokit/rest": "16.28.7",
+    "@testing-library/react": "9.1.0",
```
juanpicado: I am adding this `@testing-library/react` for experimenting using the render method.

PR #122, opened 2019-08-25
fix: remove ToReplaceByVerdaccio #108

### LINKED ISSUE #108: Font Pathing issue in Master
**Describe the bug**
I pulled the latest UI master as of today (2019-07-15) and ran the build script; then linked the ui properly up to my copy of Verdaccio.

However, several of the fonts are showing in brave as this:
`https://npm.proplugins.org/-/static/ToReplaceByVerdaccio/-/staticfonts/roboto-latin-400.woff2`
Which of course is broken url.   Coming from a CSS File:
`referer: https://npm.proplugins.org/-/static/0.style.026be7a5bc0c57143eb2.css`

   I created a symlink path on my local server to solve this issue; but whatever is creating the path to load the .woff2 files is broken.

**To Reproduce**
1. Install UI from 4.0-master in Verdaccio
2. Open dev tools; network tab.
3. Load the Verdaccio instance

**Expected behavior**
No broken links.

**Screenshots**

image

Please note the path; it is a 304; because I did create a path/symlink back to the static fonts folder; so that it can be resolved...

juanpicado:
**Type:** bug

**Description:**

it populates CSS url() props with `ToReplaceByVerdaccio` which cannot be replaced anyway, this PR remove that string.

https://github.com/verdaccio/ui/issues/108

Resolves #108

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,1 @@
 save-prefix ""
-registry "http://registry.npmjs.org/"
--- a/tools/webpack.config.js
+++ b/tools/webpack.config.js
@@ -7,7 +7,6 @@ module.exports = {
   output: {
     path: `${env.APP_ROOT}/static/`,
     filename: '[name].[hash].js',
-    // FIXME: do we need this?
     publicPath: '/-/static',
   },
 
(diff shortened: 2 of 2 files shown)


.yarnrc
```diff
@@ -1,2 +1 @@
 save-prefix ""
-registry "http://registry.npmjs.org/"
```
juanpicado: We don't need this

PR #127, opened 2019-09-11
Refactor ci

sergiohgz:
**Type:** ci

*  There is a related issue? **No**
*  Unit or Functional tests are included in the PR **An updated failing snapshot**

**Description:** Github Actions is removing support for workflows defined in HCL by the end of month. Here I remove old workflows and update Github CI config. Also, I have improved CircleCI using workspaces instead of caches, which allows external PRs to run correctly between stages of the workflow, and install dependencies from lockfile.
Finally, to get this job work in every platform, I updated the lockfile, fixed Prettier endOfLine setting and Stylelint glob expansion for running in Windows machines, and updated a failing snapshot.


juanpicado:
I think we can remove Node8 already, LTS is more than enough for this repo. So the Ci would be faster.
Screen Shot 2019-09-11 at 10 18 22 PM

sergiohgz:
Perfect, I'll drop support for Node 8 in all the repos

.circleci/config.yml
```diff
@@ -48,21 +45,11 @@ jobs:
...
-          key: *base_config_key
-          paths:
-            - ~/.npmrc
-            - ~/.gitconfig
       - restore_cache:
           key: *yarn_cache_key
       - run:
-          name: Install Js dependencies
-          command: yarn install --no-progress --registry https://registry.verdaccio.org --no-lockfile
+          name: Install dependencies
```
juanpicado: I would like to keep `--registry https://registry.verdaccio.org` otherwise will go to public in some scenarios.

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  build_test_lint:
+    name: Node ${{ matrix.node_version }} and ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node_version: ['8.16', 10, 12] # Node 8 version fix due to https://github.com/actions/setup-node/issues/27
+        os: [ubuntu-latest, windows-latest, macOS-latest]
```
juanpicado: I think ubuntu is fine or any of them, perhaps windows only, UI runs on browser anyway.

.circleci/config.yml
sergiohgz: If we use `--registry` option, we have to remove `--frozen-lockfile`, so it will slow down the installation. Maybe we should have a .yarnrc that specifies the registry to install (like in monorepo)?

.github/workflows/main.yml
sergiohgz: Yes, UI will run on browser, but contributors can develop on several different systems. I think it's not bad idea to execute CI in the three specified.

Issue #116, opened 2019-08-12
Using React Hooks (Refactoring, New Components)


juanpicado:
**Is your feature request related to a problem? Please describe.**

React Hooks is a topic we wanna to raise up in order to keep modernize the project. Currently we rely on Classes components and Stateless components.

**Describe the solution you'd like**

Since this is a new topic for the big majority, here https://github.com/verdaccio/ui/pull/115 the first step. 

**Describe alternatives you've considered**

It is not mandatory to refactor all existing components, but if someone is willing to levep up the React skills, go ahead, but here some rules for this

- If there is not test, it must included it.
- If the test already exist, you will need to either refactor it or update it.
- One component (refactoring/new) peer PR 🙏 
- All test must pass (soon also with Browserstack https://github.com/verdaccio/ui/pull/111)
- Don't be afraid to extract components if you consider the chosen one can be improved, ask for suggestions in our chat for guidance.

**Additional context**

- We have updated to React `16.9.0` so, many issues related with Enzyme now are gone. There are some limitations with `shallow` rendering, so be aware of it. https://github.com/airbnb/enzyme/issues/2011
- We added `@testing-library/react` if someone wants to experiments with it. But we keep Enzyme as a main testing tool.
- `eslint-plugin-react-hooks` was added for help linting the new components.

If you have any questions, ask http://chat.verdaccio.org `#question` channel.  Or if you have suggestions feel free to drop it here. This is a long term ticket (either called story) so, here the policy is **first come first served**, just PR and we will help with the code review.

priscilawebdev:
> We added @testing-library/react
Me when I saw it: 😍

priscilawebdev:
I will do this task 😉

gagandeepp:
@priscilawebdev  can I work on this too??

@juanpicado

priscilawebdev:
@gagandeepp yes sure... please just let me know which components you would to refactor 🙃

gagandeepp:
@priscilawebdev  I will be refactoring login module

nutboltu:
@priscilawebdev and @juanpicado Can I also work on this issue?

juanpicado:
go ahead @nutboltu 👍

agar23:
@juanpicado any components I can take up?

priscilawebdev:
@agar23 thank you for reaching out! What do you think about the component "PackageList"? it would be really nice to have it as a Functional component 😉

agar23:
[...]

PR #129, opened 2019-09-27
Version Component - Replaced classes by func. comp

priscilawebdev:
**Type:** Refactor

It was not necessary to update the existing ones

**Description:**

in order to use react hooks, in this PR I updated the Version component by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.


src/components/Versions/Versions.tsx
```diff
@@ -1,90 +1,48 @@
...
 
-import { Heading, Spacer, ListItemText } from './styles';
+import { Heading } from './styles';
+
+import VersionsTagList from './VersionsTagList';
+import VersionsHistoryList from './VersionsHistoryList';
+
+import { DIST_TAGS } from '../../../lib/constants';
 
 export const NOT_AVAILABLE = 'Not available';
```
juanpicado: Not longer use, we can remove it.

src/components/Versions/Versions.tsx
```diff
@@ -1,90 +1,48 @@
...
-            <Heading variant="subtitle1">{LABEL_VERSION_HISTORY}</Heading>
-            {this.renderPackageList(versions, timeMap, packageName)}
-          </>
-        )}
-      </>
-    );
-  }
-}
+const Versions: React.FC = () => (
+  <DetailContextConsumer>
```
juanpicado: I think you can use here https://reactjs.org/docs/hooks-reference.html#usecontext

The code will be even smaller and simpler to read.

src/components/Versions/styles.ts
```diff
@@ -14,6 +14,7 @@ export const Spacer = styled('div')({
   borderBottom: '1px dotted rgba(0, 0, 0, 0.2)',
   whiteSpace: 'nowrap',
   height: '0.5em',
+  margin: '0 16px',
```
juanpicado: Could you give more context about the margin?

priscilawebdev: I noticed that there was no space between the Tabs component and the content, so I added 16px margin. Maybe in the future we may have some common spaces, making the application more homogeneous.

src/components/Versions/Versions.tsx
priscilawebdev: oh true :-) good catch!

src/components/Versions/Versions.test.tsx
```diff
@@ -1,48 +1,39 @@
...
-  beforeEach(() => {
-    jest.resetModules();
-  });
-
   afterEach(() => {
     cleanup();
   });
 
   // FIXME: this test is not deterministic (writes `N days ago` in the snapshot, where N is random number)
   test.skip('should render the component in default state', () => {
```
juanpicado: Let's remove this in another PR.

PR #130, opened 2019-09-27
DetailContainer Component - Replaced class by func. comp 

priscilawebdev:
**Type:** Refactor

Not completely

**Description:**

in order to use react hooks, in this PR I updated the DetailContainer component by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.

I didn't add tests using the react testing library now, because the content of a tab can diverse according to the value passing from the context. I need to think how I will test it, but feedbacks are already very welcome :-)


juanpicado:
Test are failing, could you check @priscilawebdev https://app.circleci.com/jobs/github/verdaccio/ui/1522 ?

priscilawebdev:
@juanpicado the PR is a bit large now (if you prefer I can split the PR) but tests are passing 🙂

priscilawebdev:
@sergiohgz I would like to start organizing the components in order to implement our Theme...after they will be together again..i just find this way easier for now...

sergiohgz:
@priscilawebdev perfect if we use it as workaround for organizing, but I think this approach should not be permanent :wink:
Also, this PRs looks so great :wink:

juanpicado:
> @sergiohgz I would like to start organizing the components in order to implement our Theme...after they will be together again..i just find this way easier for now...

Let's make a document about how this process would take place. It's a way all we are in the same page.

https://drive.google.com/drive/folders/0B7N5pSnWAO88S0lmQUpiU0laWWM

src/components/DetailContainer/DetailContainer.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { render } from '@testing-library/react';
+
+import DetailContainer from './DetailContainer';
+
+describe('DetailContainer', () => {
+  test('renders correctly', () => {
+    const { container } = render(<DetailContainer />);
+    expect(container.firstChild).toMatchSnapshot();
+  });
+  // TODO -> implement more tests, for instance if a tab has it's correct content
```
juanpicado: Could you add with `test.todo('xxxxxx')` which scenarios would be missing? So another person can take over of it.
eg:
```
test.todo('should test click on tabs');
```

src/components/DetailContainer/DetailContainer.tsx
```diff
@@ -1,77 +1,29 @@
...
-export const UPLINKS_LABEL = 'Uplinks';
+const DetailContainer: React.FC = () => {
+  const [tabPosition, setTabPosition] = useState(TabPosition.README);
+  const detailContext = useContext(DetailContext);
+  const { readMe } = detailContext;
 
-class DetailContainer<P> extends Component<P, DetailContainerState> {
-  public state = {
-    tabPosition: 0,
+  const handleChangeTabPosition = (event: ChangeEvent<{}>) => {
```
juanpicado: Here we can use `useCallback`, so the handler is not recreated in each render.
```
  const handleChangeTabPosition = useCallback(
    (event: ChangeEvent<{}>) => {
      const eventTarget = event.target as HTMLSpanElement;
      const chosentab = eventTarget.innerText as TabPosition;
      setTabPosition(TabPosition[chosentab]);
    },
    [setTabPosition]
  );
```

juanpicado: https://reactjs.org/docs/hooks-reference.html#usecallback

juanpicado: It seems the `event.preventDefault();` was removed.

src/components/DetailContainer/DetailContainerTabs.tsx
```diff
@@ -0,0 +1,35 @@
...
+  marginBottom: 16,
+});
+
+const DetailContainerTabs: React.FC<Props> = ({ tabPosition, onChangeTabPosition }) => {
+  const [tabPositionIndex, setTabPositionIndex] = useState(0);
+
+  useEffect(() => {
+    const tabIndex = Object.keys(TabPosition).findIndex(position => position === String(tabPosition).toUpperCase());
+    setTabPositionIndex(tabIndex);
+  }, [tabPosition]);
```
juanpicado: I think also `setTabPositionIndex` should be part of the list.

https://reactjs.org/docs/hooks-reference.html#useeffect

>  every value referenced inside the effect function should also appear in the dependencies array. In the future, a sufficiently advanced compiler could create this array automatically

src/components/DetailContainer/types.ts
```diff
@@ -1,7 +0,0 @@
-import { ReactNode } from 'react';
```
juanpicado: This is the better approach, keep types with each component, less files to lint, faster builds. As less files we have, the better.

src/components/DetailContainer/DetailContainerTabs.tsx
juanpicado: You can have `Object.keys(TabPosition)` outside of the component, it will never change anyway, this suggestion  is mostly an optimization.

src/components/DetailContainer/DetailContainer.tsx
priscilawebdev: Oh it's true...I have updated the code, now it uses useCallback and event. preventDefault( thank you!!

src/components/DetailContainer/types.ts
priscilawebdev: I am not sure if I understood you ... did you like my changes?

src/components/DetailContainer/DetailContainerTabs.tsx
priscilawebdev: Hmmm are you sure? because the `setTabPositionIndex` won't change.

priscilawebdev: I checked twice now and couldn't find any examples, also react official, having the setFunction in the array

src/components/DetailContainer/types.ts
juanpicado: Yes, my point is, having less file, speed up the whole build. Less I/O, so, cool to see less file 😎

src/components/DetailContainer/DetailContainerTabs.tsx
juanpicado: https://stackoverflow.com/a/55743744 👍

PR #141, opened 2019-09-29
fix: cli command

kopax:
**Type:**

*  There is a related issue: https://github.com/verdaccio/verdaccio/issues/1481#issuecomment-536233237
*  Unit or Functional tests are included in the PR: nop

**Description:**

When JWT is used, `--always-auth` should always be used, as stated in [npm `cli` documentation](https://docs.npmjs.com/cli/adduser#always-auth):

> ### always-auth
> 
> Default: false
> 
> If specified, save configuration indicating that all requests to the given 
> registry should include authorization information. Useful for private
> registries. Can be used with `--registry` and / or `--scope`, e.g.
> 
>     npm adduser --registry=http://private-registry.example.com --always-auth

> This will ensure that all requests to that registry (including for tarballs)
> include an authorization header. This setting may be necessary for use with
> private registries where metadata and package tarballs are stored on hosts with
> different hostnames. See `always-auth` in `npm-config(7)` for more details on
> always-auth. Registry-specific configuration of `always-auth` takes precedence
> over any global configuration.

Better always showing it even with legacy security than not showing it, this is displayed to all my users as a **helper** and can conduct to time lost for many of them.

### DIFF
--- a/src/components/Help/Help.tsx
+++ b/src/components/Help/Help.tsx
@@ -30,7 +30,7 @@ const Help: React.FC = () => {
         <HelpTitle color="textSecondary" gutterBottom={true}>
           {'To publish your first package just:'}
         </HelpTitle>
-        {renderHeadingClipboardSegments('1. Login', `npm adduser --registry ${registryUrl}`)}
+        {renderHeadingClipboardSegments('1. Login', `npm adduser --registry ${registryUrl} --always-auth`)}
         {renderHeadingClipboardSegments('2. Publish', `npm publish --registry ${registryUrl}`)}
         <Typography variant="body2">{'3. Refresh this page.'}</Typography>
       </CardContent>
--- a/src/utils/cli-utils.ts
+++ b/src/utils/cli-utils.ts
@@ -23,7 +23,7 @@ export function getCLISetConfigRegistry(command: string, scope: string, registry
 }
 
 export function getCLISetRegistry(command: string, registryUrl: string): string {
-  return `${command} --registry ${registryUrl}`;
+  return `${command} --registry ${registryUrl} --always-auth`;
 }
 
 export function getCLIChangePassword(command: string, registryUrl: string): string {


sergiohgz:
I am using JWT and didn't use or set always-auth flag. I will dig into this when I have time.
Also, the snapshot tests were not updated with the changes.
Last thing, the flag will appear when using legacy authentication, not only when using JWT, so that is incorrect.

kopax:
> I am using JWT and didn't use or set always-auth flag. I will dig into this when I have time.

Surprising but not impossible. I was working without it in development but as soon as I went to production, I lost hours digging for the issue and that was it.

> Also, the snapshot tests were not updated with the changes.

I will have a look.

> Last thing, the flag will appear when using legacy authentication, not only when using JWT, so that is incorrect.

It's better to have it always than always not having it, because it won't add any header nor break the authentication to anyone.

juanpicado:
> I am not using snapshot, what are they for? how do you update them?

https://jestjs.io/docs/en/snapshot-testing

sergiohgz:
I reviewed several scenarios, with registry installed on different hosts, with specific auth plugins for all of them, and my conclusion is that this is not necessary. As the [npm cli docs](https://docs.npmjs.com/cli/adduser#always-auth) said, it may be necessary for private registry, but not in all cases.

kopax:
> https://jestjs.io/docs/en/snapshot-testing

Thanks, I will have a look until we find an acceptable solution.

> I reviewed several scenarios, with registry installed on different hosts, with specific auth plugins for all of them, and my conclusion is that this is not necessary. As the [npm cli docs](https://docs.npmjs.com/cli/adduser#always-auth) said, it may be necessary for private registry, but not in all cases.

It may be necessary and it's unclear when and why. As a result, some repository will display a wrong helper to all their users, this can result in more confusion. 

A solution for theim must be done.

sergiohgz:
The own npm documentation is not clear when they are needed (I don't see any use case when I need it) and I think it's not Verdaccio responsability educate developers about npm cli tool. 
Can you open an issue with an example demostrating its need in legacy and JWT modes and steps to reproduce?

kopax:
> The own npm documentation is not clear when they are needed

Agree.

> I don't see any use case when I need it

I can give you one, I can give you access to my verdaccio production server so you can see a case where you MUST have it.

Also see for example:

- https://github.com/verdaccio/verdaccio/issues/1410#issuecomment-515703763
- https://github.com/verdaccio/verdaccio/issues/1462#issuecomment-529780065
- https://github.com/verdaccio/verdaccio/issues/1381#issuecomment-511699858
- https://github.com/verdaccio/verdaccio/issues/1381#issuecomment-536245726
- https://github.com/verdaccio/verdaccio/issues/1381#issuecomment-536285874
- https://github.com/verdaccio/verdaccio/issues/1105#issuecomment-435936158
- https://github.com/verdaccio/verdaccio/issues/1088#issuecomment-432225622
- https://github.com/verdaccio/verdaccio/issues/999#issuecomment-421638649
- https://github.com/verdaccio/verdaccio/issues/978#issuecomment-418800115

>  I think it's not Verdaccio responsability educate developers about npm cli tool.

We are giving wrong helper to some of them, it's kind of verdaccio responsibility. We should at least warn them, at least a small info. Again, this was happening on all the clients, me, CIs. It even got recommended by many verdaccio core members.

> Can you open an issue with an example demostrating its need in legacy and JWT modes and steps to reproduce?

No, I'd like to but I can't, as I told you, I have hitten the problem in production not in development,  both configurations were identical.

sergiohgz:
I have several production servers to test if you give a way to reproduce. Without a way to repro it's difficult to see specific configurations and logs of what is happening.

Seeing all the comments you mention above, I think that it could be more a documentation troubleshoot rather than a forced option or advise in the main UI for all clients. As I said before, not every users need this option and it could create some noise in their application UIs
What do you think about a troubleshooting section in website with some client logs?

juanpicado:
> We are giving wrong helpers to some of them, it's kind of verdaccio responsibility, otherwise, remove them (I am not even telling you that all those utilities are also wrong if you see them from a package page instead of the home page). We should at least warn them, at least a small info tips.

Agree, but not in the UI. That might lead to miss-leading information, if tomorrow `npm` changes the rules, I'd have thousands of users asking me why is not working, it just impossible to teach everybody via UI. I'd take the @sergiohgz recommendation, that new section would be much helpful, due the website is the root of knowledge for everybody and can scale to multiple topics in the future and capable to be translated.

kopax:
Fair enough. Thanks for the understanding effort. If it's that important we'll probably hear more about it soon.

PR #142, opened 2019-10-01
Header Component - Replaced class by func. comp

priscilawebdev:
**Type:** Refactor

Yes. 

**Description:**

in order to use react hooks, in this PR I updated the Header component by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.


juanpicado:
Your last PR conflict with this one https://github.com/verdaccio/ui/commit/f84fd79c5b37f27f226acadaca4c3c81eb2e4fb8#diff-2780c4d0882ad10abae71aa33af34c7b

juanpicado:
+1% coverage 🥰 good job 👏🏻

src/components/Header/Header.test.tsx
```diff
@@ -1,124 +1,138 @@
...
-    const state = {
-      openInfoDialog: false,
-      packages: undefined,
-      registryUrl: 'http://localhost',
-      showMobileNavBar: false,
-    };
-    expect(wrapper.state()).toEqual(state);
-    expect(routerWrapper.html()).toMatchSnapshot();
+    const documentationBtn = getByTestId('header--tooltip-documentation');
+    fireEvent.click(documentationBtn);
```
juanpicado: I guess you need to check whether button was clicked right?

src/components/Header/Header.test.tsx
```diff
@@ -1,125 +1,138 @@
...
-    routerWrapper = shallow(
+  test('should load the component in logged out state', () => {
+    const { container, queryByTestId, getByText } = render(
+      <Router>
+        <Header onLogout={headerProps.handleLogout} onToggleLoginModal={headerProps.handleToggleLoginModal} scope={headerProps.scope} />
+      </Router>
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
+    expect(queryByTestId('header--menu-acountcircle')).toBeNull();
```
juanpicado: Reminder, the test-id is fine for me, but, according the guidelines, should be the 3rd option. 

https://testing-library.com/docs/guide-which-query

aforty:
This commit had the unintended consequence of breaking the plugin `verdaccio-github-oauth-ui` which was relying on `id="#header--button-login"` and `id="#header--button-logout"` being set on the Login and Logout buttons.

Not sure what the fix here is. Add the id's back to modify thew plugin to look for the id's in `data-testid`?

Issue to track is here: https://github.com/n4bb12/verdaccio-github-oauth-ui/issues/39

PR #145, opened 2019-10-01
Upgraded Typescript to 3.6.3

tmkn:
This PR updates TypeScript to 3.6.3.

The improved typechecking of 3.6 highlighted 2 issues.
1 was easy to fix by using built in types, the 2nd one was a little more delicate as the material ui typings are not great/wrong.

Namely the `withWidth` hoc. Due to the way the typings where written they show up as optional properties on the component itself. So when you instantiate the component you can optionally pass in a `width` prop. Which of course doesn't make sense since it is injected by the hoc automatically.

This behaviour makes it so that inside your component `width` has the type `Breakpoint | undefined` ._.
As such you can't directly pass it on to `isWidthUp('sm', width)` as it expects `Breakpoint` but as said the hoc prop is (incorrectly) typed as `Breakpoint | undefined`. This needs to be fixed in material ui.

So for the time being I had to overrule the compiler and disable the eslint no null assertion test for this line to fix the wrong hoc `width` prop type from material ui: `sWidthUp('sm', width!)`.
Since I saw other places where eslint rules where occasionally disabled I felt confident to also do it here rather than to do an explicit null check in the code (Since there is nothing wrong with the actual code to begin with).

I also needed to  add `@typescript-eslint/parser` as the parser for `eslint`
The default parser(espree) only supports `typescript` < 3.6. . `@typescript-eslint/parser` in the meantime is now the preferred parser for `typescript`.

e: I also added a setting for vscode so that it uses the typescript version from the package.json rather than the built in from vscode.


juanpicado:
Great explanation @tmkn

PR #147, opened 2019-10-02
build: use current package manager

DanielRuf:
**Type:**
build

**Description:**

This PR uses the environment setting to use the current package manager for the commands.


DanielRuf:
We will have to test all commands on all three / two platforms: Unix / Linux + macOS and Windows.

sergiohgz:
Can you try this in monorepo? There we have CI in Windows, Linux and MacOS, so it could be easier to test on all main systems

sergiohgz:
I tried the use of $npm_execpath on lint stage in [monorepo](https://github.com/verdaccio/monorepo/compare/use-npm-execpath-pm) and you can see that it won't work on Windows. Check results for [Node 12](https://github.com/verdaccio/monorepo/runs/244651077) and in [Node 10](https://github.com/verdaccio/monorepo/runs/244651023), the variable is not found.

DanielRuf:
@sergiohgz hm, can you try `cross-env-shell` instead?

See https://github.com/kentcdodds/cross-env/issues/91

DanielRuf:
I will squash them then when this resolves it.

sergiohgz:
https://github.com/verdaccio/monorepo/runs/244666867

At the moment of writing this comment, it has been stuck for 10 minutes in the step, it seems to not work :disappointed:

sergiohgz:
If I understand this section of [`cross-env` docs](https://github.com/kentcdodds/cross-env#windows-issues), `cmd` doesn't accept command substitution

DanielRuf:
> If I understand this section of [`cross-env` docs](https://github.com/kentcdodds/cross-env#windows-issues), `cmd` doesn't accept command substitution

But `cross-env-shell` should be able to do this.
I have to test this on my own Winbdows machine when I am at home. But I use Git Bash there. I might have to check if my Windows cmd works there.

DanielRuf:
> If I understand this section of [`cross-env` docs](https://github.com/kentcdodds/cross-env#windows-issues), `cmd` doesn't accept command substitution

Hm, weird. I thought this worked at home in another project.
But `cross-env-shell` should theoretically be able to do this.

I have to test this on my own Winbdows machine when I am at home. But I use Git Bash there. I might have to check if my Windows cmd works there.

sergiohgz:
Basically, it suggest to use Powershell to fix the problem. Should we enforce users to use a specific terminal? I don't think so.
In my case, I use VMs or WSL on Windows, so I cannot test it directly

DanielRuf:
Right. I'm closing this PR then. I thought it works as we use it in another project too.

juanpicado:
Having fun without me 🤓  .... bad boys ...

sergiohgz:
@juanpicado hahaha you know 😜

PR #148, opened 2019-10-02
fix: warning about modules with names differing in casing

antoinechalifour:
**Type:**

_The following has been addressed in the PR:_

This PR removes a warning in the dev environment.

**Description:**

The `module.accept` method was called with a different casing with resulted in a warning in the console :)

image


priscilawebdev:
Thank you @antoinechalifour 😊

PR #150, opened 2019-10-02
refactor: convert Author component to hooks

bighuggies:
**Type:** refactor

Minor changes had to be made to the mocking in the existing tests to enable hooks. The tests are also more typesafe.

**Description:**
- Convert `<Author />` to a function component
- Replace render prop context consumer with `useContext` hook
- Removed some indirection in the JSX by inlining functions
- Add missing types to `packageMeta` (is there a better way to do this?)


src/components/Author/Author.test.tsx
```diff
@@ -51,14 +41,13 @@ describe('<Author /> component', () => {
       latest: {
         name: 'verdaccio',
         version: '4.0.0',
+        dist: { fileCount: 0, unpackedSize: 0 },
       },
+      _uplinks: {},
```
bighuggies: These are not needed for the test but were missing before. The type error was hidden by using `jest.fn`

src/components/Author/Author.tsx
```diff
@@ -1,58 +1,44 @@
...
-    return (
-      <a href={`mailto:${email}?subject=${packageName}@${version}`} target={'_top'}>
-        {avatarComponent}
-      </a>
-    );
+  if (!author) {
+    return null;
   }
 
-  public renderAuthor = ({ latest }) => {
```
bighuggies: `latest` was implicitly typed as `any` before - this method wasn't type safe and was accessing properties that weren't defined on the type of `packageMeta`

src/components/Author/Author.test.tsx
```diff
@@ -51,14 +41,13 @@ describe('<Author /> component', () => {
...
     };
 
-    // @ts-ignore
-    mockPackageMeta.mockImplementation(() => packageMeta);
-
-    const wrapper = mount(<Authors />);
-    expect(wrapper.html()).toEqual('');
+    const wrapper = mount(component(packageMeta));
+    expect(wrapper.html()).toBeNull();
```
bighuggies: Previously the context `Consumer` was always rendered so this was an empty string. It changed to `null` since now if the context value is empty we return `null` (and don't have to render the `Consumer`).

bighuggies:
Tests failing - I think it's a platform or version issue (I'm on Windows). Running `yarn install && yarn test` on `master` fails on my machine due to different generated classes.

src/components/Author/Author.tsx
```diff
@@ -1,58 +1,44 @@
...
-}
+  const { email, name } = author;
+
+  const avatarComponent = <Avatar alt={author.name} src={author.avatar} />;
+
+  return (
+    <List subheader={<Heading variant={'subtitle1'}>{'Author'}</Heading>}>
+      <AuthorListItem button={true}>
+        {!email || !isEmail(email) ? (
+          <>{avatarComponent}</>
```
priscilawebdev: here you don't need to wrapper the const with fragments , you can just do like this:

```suggestion
       avatarComponent
```

src/components/Author/Author.test.tsx
```diff
@@ -1,30 +1,21 @@
...
-  },
-}));
+import Authors from './Author';
 
 describe('<Author /> component', () => {
   beforeEach(() => {
     jest.resetAllMocks();
   });
 
+  const component = (packageMeta: React.ContextType<typeof DetailContext>['packageMeta']): JSX.Element => (
```
priscilawebdev: Hmmm, as it is a component, it must be capitalize and have a more descriptive name

priscilawebdev: I would also write this component outside of the describe<'Author .....

types/packageMeta.ts
```diff
@@ -1,11 +1,19 @@
+export interface Author {
+  name: string;
+  avatar?: string;
+  email?: string;
+}
+
 export interface PackageMetaInterface {
   latest: {
+    author?: Author;
```
priscilawebdev: Author can be either a string or an object

src/components/Package/Package.tsx
```diff
@@ -45,7 +40,7 @@ export interface PackageInterface {
   name: string;
   version: string;
   time?: number | string;
-  author: Author;
+  author: Required<PackageMetaInterface['latest']>['author'];
```
priscilawebdev: Hmmm I would rather use the Author interface + string here.

priscilawebdev:
@bighuggies PR looks great! Thank you so much 👏, it needs just some minor changes..please check my comments.

src/components/Author/Author.tsx
juanpicado: You missed the `{ }` curly brackets ...

src/components/Author/Author.test.tsx
juanpicado: `component` ==> `mockAuthorComponent` . It has to be lower case since is a function. Other popular naming way might be `withAuthorComponent`.

types/packageMeta.ts
juanpicado: Yes, the metadata is quite flexible, https://docs.npmjs.com/files/package.json#people-fields-author-contributors you can see the different shape of `Author`. 

Here an example. https://github.com/verdaccio/monorepo/blob/master/core/types/index.d.ts#L39

You might use `import {Author} from `@verdaccio/types` as well, so  you don't have to duplicate the types.

src/components/Package/Package.tsx
juanpicado: I will resolve this one, above is already explained.

src/components/Author/Author.test.tsx
juanpicado: You might use https://github.com/testing-library/react-testing-library/  if you feel more confortable, as `enzyme` is also available. https://github.com/testing-library/react-testing-library/

bighuggies: I did initially rewrite these with rtl. I stuck with enzyme in the end so the snapshots would still match in this PR and it would be easier to see that there are no regressions in the test cases.

types/packageMeta.ts
bighuggies: I had a go using the type from the types package - unfortunately it's missing the `avatar` property so it causes more type errors.

I also tried rewriting as `string | Author` - it also causes a number of type errors (e.g. the `Author` component assumes `author` is an object and doesn't handle parsing the string version).

Since it's an existing typing bug, would you be OK considering it outside the scope of this PR?

src/components/Package/Package.tsx
bighuggies: I used this gross type to avoid renaming the `Author` type import to fix the conflict with the `Author` styled component because I couldn't think of something to rename it to. I went with `PackageAuthor` - not ideal but maybe a bit better than this

types/packageMeta.ts
juanpicado: Yes, make it green and we will fix it later. I will write it down. Thanks for the research.

src/components/Author/Author.test.tsx
priscilawebdev: Hmm unfortunately I don't agree with you @juanpicado it returns an JSX element and according to the standards it should be capitalize...Please see that also Function Components are functions

bighuggies: I think the main difference is that this doesn't take props, you can't use it as a JSX element.

That's why I didn't consider it a component and why I used lowercase, it's more of a function that returns JSX.

bighuggies:
That seems to be an issue on `master` as well.

A quick `git bisect` finds this change is probably responsible: https://github.com/verdaccio/ui/commit/f84fd79c5b37f27f226acadaca4c3c81eb2e4fb8?w=1#diff-8b0edc064390f472a826cacf9ea97dc4R6

I will open a separate PR to fix it 👍

juanpicado:
Thanks @bighuggies

src/components/Author/Author.test.tsx
juanpicado: @priscilawebdev any last comment here? or can we merge?

priscilawebdev: @juanpicado no...We can merge :-)

priscilawebdev:
thank you @bighuggies... yeah the Tooltip import was wrong.. I need to check why the tests passed...good job!

PR #152, opened 2019-10-03
Install Component - Replaced class by func. comp

priscilawebdev:
**Type:** Refactor

Yes.

**Description:**

in order to use react hooks, in this PR I updated the Install component by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.


src/components/Install/Install.test.tsx
```diff
@@ -1,11 +1,48 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
+
+import { DetailContext, DetailContextProps } from '../../pages/Version';
+import data from '../Versions/__partials__/data.json';
```
juanpicado: Better duplicate than reference other component resources, if someone needs to change the metadata or make it smaller will break your test. The better here is create your own `__partials__`, isolated test resources helps with maintenance in long term.

priscilawebdev:
@juanpicado good catch!! it should be fine now :-)

PR #155, opened 2019-10-03
Fix some warnings in console

antoinechalifour:
**Type:** Fix

**Description:**

Fixes 2 warnings in the console:

- unknown react emotion selector
- DOM nesting validation (`div` cannot be a child of `p`)


src/components/DetailSidebar/DetailSidebar.tsx
```diff
@@ -14,14 +14,18 @@ import Repository from '../Repository/Repository';
...
+import { TitleListItem, TitleListItemText, PackageDescription, PackageVersion } from './styles';
 
 const renderLatestDescription = (description, version, isLatest: boolean = true): JSX.Element => {
   return (
-    <span>
-      <div>{description}</div>
-      {version ? <small>{`${isLatest ? 'Latest v' : 'v'}${version}`}</small> : null}
-    </span>
+    <>
+      <PackageDescription>{description}</PackageDescription>
```
juanpicado: Coud you give me more context about this change? 🤔

src/components/DetailSidebar/styles.ts
```diff
@@ -21,10 +18,14 @@ export const TitleListItemText = styled(ListItemText)({
   },
 });
 
-export const TitleAvatar = styled(Avatar)({
+export const PackageDescription = styled('span')({
+  '&&': {
```
juanpicado: `<div>` is block by default, you might switch to `<div>` to avoid this CSS. Somehow, I guess this is related with my comment above. So, l wait for the context.

src/components/DetailSidebar/DetailSidebar.tsx
antoinechalifour: `div` is not a valid decendant of `p`, thus a warning in  the console. I refactored the DOM to be a `span` with a block display instead of a `div`

src/components/DetailSidebar/styles.ts
antoinechalifour: It is !
See my comment above :)

src/components/DetailSidebar/DetailSidebar.tsx
juanpicado: Thanks for the clarification 🚀

PR #156, opened 2019-10-03
refactor: convert Dist component to hooks

bighuggies:
**Type:** refactor

*  Unit or Functional tests are included in the PR - yes, similar minor changes to #150 

**Description:**
Very similar to #150 

- Convert `Dist` component to a function component
- Convert `DetailContext` consumer to `useContext` hook
- Inline render methods to simplify JSX


src/components/Dist/Dist.tsx
```diff
@@ -1,56 +1,43 @@
...
-
-  private renderChips(dist, license: PackageMetaInterface['latest']['license']): (JSX.Element | undefined)[] {
-    const distDict = {
-      'file-count': dist.fileCount,
-      size: dist.unpackedSize && fileSizeSI(dist.unpackedSize),
-      license,
-    };
-
-    const chipsList = Object.keys(distDict).map((dist, key) => {
-      if (!distDict[dist]) return;
```
bighuggies: Not sure if this is a bug or intentional, but this will hide the chips for file count or size if they happen to be 0. I've assumed it's intentional and included the same behavior.

PR #157, opened 2019-10-04
Improved typings

tmkn:
This PR adds type information to libraries that didn't previously had types, specifically
- autosuggest-highlight
- js-base64
- localstorage-memory

`js-base64` was incorrectly used in a unit test, I adapted the test.
I also needed to changed the signature of `isTokenExpire(token: string | undefined)` to `isTokenExpire(token: string | null)` since `localStorage.getItem` actually returns `string | null` and not `string | undefined`


src/utils/login.test.ts
```diff
@@ -33,11 +34,16 @@ describe('isTokenExpire', (): void => {
...
   test('isTokenExpire - token is not a valid json token', (): void => {
-    const token = generateTokenWithExpirationAsString();
-    const result = ['Invalid token:', new SyntaxError('Unexpected token o in JSON at position 1'), 'xxxxxx.W29iamVjdCBPYmplY3Rd.xxxxxx'];
+    const token = generateInvalidToken();
+    const result = ['Invalid token:', new SyntaxError('Unexpected token i in JSON at position 0'), 'xxxxxx.aW52YWxpZHRva2Vu.xxxxxx'];
     expect(isTokenExpire(token)).toBeTruthy();
     expect(console.error).toHaveBeenCalledWith(...result);
   });
+
+  test('isTokenExpire - token expiration is not a number', (): void => {
```
tmkn: I split the old unit test into 2, one that checks for an invalid token and one that checks for the expiration being a string

priscilawebdev:
@tmkn Great PR! Danke Thomas 🙃

PR #158, opened 2019-10-05
Use React.lazy for loading components

antoinechalifour:
**Type:** Refactor

The following has been addressed in the PR: Use `React.Lazy` for loading components

*  There is a related issue? Nop
*  Unit or Functional tests are included in the PR None changed


juanpicado:
Could you share any analysis that prove the impact of this changes in the bundle size? chunks etc etc ?

antoinechalifour:
I'll try to give you some numbers tomorrow but here's the short version :

- using `React.lazy` is the idiomatic way of doing code splitting in react land
- it removed the `asyncComponent` file so the bundle should be lighter

priscilawebdev:
@antoinechalifour Well done! 👏 Maybe you could also remove the file "asynComponent", because it is not being used anymore

src/router.tsx
```diff
@@ -25,7 +24,7 @@ class RouterApp extends Component<RouterAppProps> {
   public render(): ReactElement<HTMLDivElement> {
     return (
       <Router history={history}>
-        <>
+        <React.Suspense fallback={null}>
```
priscilawebdev: Please add the loading component here
```suggestion
        <React.Suspense fallback={<Loading />}>
```

antoinechalifour: For now the `asyncComponent` renders `null` so I wanted to keep the existing behavior. But I can change it to display a loading component

antoinechalifour:
I think it's already been removed ! I'll check tomorrow

src/router.tsx
juanpicado: `<Loading/>` already are defined in each page, we can skip this.

PR #159, opened 2019-10-05
Routes - Replaced class by func. comp

priscilawebdev:
**Type:** Refactor

 in progress

**Description:**

in order to use react hooks, in this PR I updated the routes by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.


tslint.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": [
```
juanpicado: We do not use tslint anymore @priscilawebdev

priscilawebdev: I completely forgot, but then I remembered ... so I didn't finish

src/routers.ts
```diff
@@ -0,0 +1,7 @@
+export enum Route {
+  ROOT = '/',
```
juanpicado: Let's keep this within the route file, they won't be used anywhere and not  change to be reused .

src/utils/package.ts
```diff
@@ -57,10 +57,6 @@ export function formatDateDistance(lastUpdate): string {
   return distanceInWordsToNow(new Date(lastUpdate));
 }
 
-export function buildScopePackage(scope: string, packageName: string): string {
-  return `@${scope}/${packageName}`;
```
juanpicado: We need this method.

src/App/App.tsx
```diff
@@ -85,6 +85,7 @@ export default class App extends Component<{}, AppStateInterface> {
   public loadOnHandler = async () => {
     try {
       const packages = await API.request<any[]>('packages', 'GET');
+      // @ts-ignore: FIX THIS TYPE:  Type 'any[]' is not assignable to type '[]'
```
juanpicado: I think `master` will solve this.

src/utils/package.ts
priscilawebdev: Hmmm it's not being used anywhere else..only in the Version Page

src/App/App.tsx
priscilawebdev: unfortunately it wasn't :/

src/App/App.tsx
```diff
@@ -1,47 +1,23 @@
...
-  showLoginModal: boolean;
-  isLoading: boolean;
-}
-export default class App extends Component<{}, AppStateInterface> {
-  public state: AppStateInterface = {
+
+import AppRoute from './AppRoute';
+import { AppProps, AppContextProvider } from './AppContext';
+
+export default class App extends Component<{}, AppProps> {
```
tmkn: The `AppProps` name is a little misleading here. It's used for the state of a class component but its name is `AppProps` :/ It's not immediately clear that this is on purpose.
I would suggest renaming the import
```typescript
import { AppProps as AppState, AppContextProvider } from './AppContext';
```

priscilawebdev: Good point Tom, but as I am working on the refactor of the App Component, I believe we can merge like it is now 😉

juanpicado: Ok @priscilawebdev , it's fine, but I will memorize this :P later on.

Issue #162, opened 2019-10-06
ForwardRef warnings


priscilawebdev:
**Current behaviour:**
As we are using a Material UI version above v3 and not all of our components in conjunction with Material-UI are not able to pass down the reference of parents to their children, we are  getting some warnings about ForwardRef. 

Example:

> Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

**Expected behaviour:**

We should wrap all Material UI components with ForwardRef, in order to fix this issue. You can read more about here: https://material-ui.com/guides/composition/#caveat-with-refs and here https://github.com/mui-org/material-ui/issues/14415

juanpicado:
Great research.

FilipMessa:
Do you need `forwardRef` for all components?
I didn't see that you use ref in all components from the `material-ui`  What is the reason for that?  Maybe I missed something.

BTW:  Does anybody know why I have few lint errors in my branch related to `./muiComponents` folder? (https://github.com/verdaccio/ui/pull/189)

Thanks, I am a beginner with typescript so any Idea will be welcome.

juanpicado:
hi @FilipMessa, please double check you have all changes from master, if the issue persist I will take care of it.

FilipMessa:
@juanpicado I used a rebase, so everything should be updated, but the problem still persists.
I have a plan to find some time and look at that today again but  I appreciate any help. 😉
thanks

juanpicado:
I will take a look no worries, for now just use `--no-verify` on commit so do not block you if you wan to push changes.

priscilawebdev:
I believe we can already close this one 👏🙂

PR #163, opened 2019-10-06
Avatar Component - Introduced ForwardRef

priscilawebdev:
**Type:**  Refactor

**The following has been addressed in the PR:** https://github.com/verdaccio/ui/issues/162

**Unit or Functional tests are included in the PR?** Not yet

**Description:**

Wrapped Material UI Avatar component with ForwardRef


juanpicado:
@priscilawebdev conflict ⬆️

src/muiComponents/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { forwardRef } from 'react';
+import { default as MaterialUIAvatar, AvatarProps } from '@material-ui/core/Avatar';
+
+// The default element type of MUI's Avatar is 'div' and we don't allow the change of this prop
+type AvatarRef = HTMLElementTagNameMap['div'];
+
+/* eslint-disable verdaccio/jsx-spread */
```
juanpicado: I see this pattern being repeated in all `muiComponents`, thus, we can create a `.eslintrc` file at ` src/muiComponents/` level and disable those rules for all files within that folder.

eg: https://github.com/verdaccio/ui/blob/master/test/acceptance/.eslintrc

priscilawebdev: Good idea @juanpicado .. I will do that 🙂

PR #167, opened 2019-10-07
Added eslint-plugin-import

priscilawebdev:
**Type:** Feat?

**The following has been addressed in the PR:**

There is no issue for this PR

**Description:**

In this PR, I am proposing to use the plugin eslint-plugin-import to enforce a convention in module import order. You can read more about it here: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

Please what do you think?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,6 +3,7 @@ coverage/
 static/
 .github/
 .circleci/
+build
 *.md
 *.lock
 *.yaml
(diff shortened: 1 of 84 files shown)


priscilawebdev:
I am glad you liked it @sergiohgz  sure i will do that 😉🙃

Issue #168, opened 2019-10-08
Disable eslint rules


antoinechalifour:
**Is your feature request related to a problem? Please describe.**
Some Eslint rules are really strict and are actually considered anti-pattern (for instance `eslint(react/jsx-no-bind)`).

Similarly :
- I don't see why `react/jsx-no-literals` is helping with the code base, its adding unnecessary syntax that is not really idiomatic to any other React codebase
- `react/destructuring-assignment` is really verbose for reading nested properties
- many other rules are not really standard

**Describe the solution you'd like**
We should consider disabling them and stick to `eslint-config-react-app` (the set of rules provided by the React team with create-react-app).

sergiohgz:
@antoinechalifour thanks for the points, are really helpful to improve our ESLint config. I am rethinking all the linting rules using our base [@verdaccio/eslint-config](https://github.com/verdaccio/monorepo/tree/master/tools/eslint-config) and the [fixed config for UI](https://github.com/verdaccio/ui/tree/master/.eslintrc). Stay tuned for more :wink:

juanpicado:
> eslint(react/jsx-no-bind

We need to get rid of them, I guess only few of them remains.

> react/jsx-no-literals

The reason behind of it is https://github.com/verdaccio/ui/issues/102 which I want to introduce it soon.

> react/destructuring-assignment

Could you give me an example @antoinechalifour ?

> many other rules are not really standard

Let's share them, perhaps I'm not aware of it. 🙃

DanielRuf:
Not sure if we get any further feedback here and if we should change something.

PR #169, opened 2019-10-08
Refactor: Dependencies - Replaced with func. comp

antoinechalifour:
**Type:** refactor

*  There is a related issue? #116 
*  Unit or Functional tests are included in the PR Yes

**Description:**

Moves `<Dependencies />` to Function Component syntax.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/node": "12.7.8",
     "@types/react": "16.9.2",
     "@types/react-dom": "16.9.0",
-    "@types/react-router-dom": "4.3.5",
+    "@types/react-router-dom": "5.1.0",
     "@types/validator": "10.11.3",
     "@typescript-eslint/parser": "2.3.2",
     "@verdaccio/babel-preset": "2.0.0",
@@ -87,8 +87,6 @@
     "react-dom": "16.10.0",
     "react-emotion": "9.2.12",
     "react-hot-loader": "4.12.11",
-    "react-router": "5.0.1",
-    "react-router-dom": "5.0.1",
     "resolve-url-loader": "3.1.0",
     "rimraf": "3.0.0",
     "source-map-loader": "0.2.4",
@@ -208,5 +206,8 @@
     "type": "opencollective",
     "url": "https://opencollective.com/verdaccio",
     "logo": "https://opencollective.com/verdaccio/logo.txt"
+  },
+  "dependencies": {
+    "react-router-dom": "5.1.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "react": "16.10.0",
     "react-autosuggest": "9.4.3",
     "react-dom": "16.10.0",
+    "react-router-dom": "5.1.2",
     "react-emotion": "9.2.12",
     "react-hot-loader": "4.12.11",
     "resolve-url-loader": "3.1.0",
@@ -206,8 +207,5 @@
     "type": "opencollective",
     "url": "https://opencollective.com/verdaccio",
     "logo": "https://opencollective.com/verdaccio/logo.txt"
-  },
-  "dependencies": {
-    "react-router-dom": "5.1.2"
   }
 }
--- a/src/components/Dependencies/Dependencies.test.tsx
+++ b/src/components/Dependencies/Dependencies.test.tsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { HashRouter } from 'react-router-dom';
+
+import { DetailContextProvider } from '../../pages/Version';
+import Dependencies from './Dependencies';
+
+describe('<Dependencies /> component', () => {
+  test('Renders a message when there are no dependencies', () => {
+    // Given
+    const packageMeta = {
+      latest: {
+        name: 'verdaccio',
+        version: '4.0.0',
+        author: {
+          name: 'verdaccio user',
+          email: 'verdaccio.user@verdaccio.org',
+          url: '',
+          avatar: 'https://www.gravatar.com/avatar/000000',
+        },
+        dist: { fileCount: 0, unpackedSize: 0 },
+        dependencies: {},
+        devDependencies: {},
+        peerDependencies: {},
+      },
+      _uplinks: {},
+    };
+
+    // When
+    const { getByText } = render(
+      <DetailContextProvider value={{ packageMeta }}>
+        <Dependencies />
+      </DetailContextProvider>
+    );
+
+    // Then
+    expect(getByText('verdaccio has no dependencies.')).toBeDefined();
+  });
+
+  test('Renders a link to each dependency', () => {
+    // Given
+    const packageMeta = {
+      latest: {
+        name: 'verdaccio',
+        version: '4.0.0',
+        author: {
+          name: 'verdaccio user',
+          email: 'verdaccio.user@verdaccio.org',
+          url: '',
+          avatar: 'https://www.gravatar.com/avatar/000000',
+        },
+        dist: { fileCount: 0, unpackedSize: 0 },
+        dependencies: {
+          react: '16.9.0',
+          'react-dom': '16.9.0',
+        },
+        devDependencies: {
+          'babel-core': '7.0.0-beta6',
+        },
+        peerDependencies: {
+          'styled-components': '5.0.0',
+        },
+      },
+      _uplinks: {},
+    };
+
+    // When
+    const { getByText } = render(
+      <HashRouter>
+        <DetailContextProvider value={{ packageMeta }}>
+          <Dependencies />
+        </DetailContextProvider>
+      </HashRouter>
+    );
+
+    // Then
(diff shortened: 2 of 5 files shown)


src/components/Dependencies/Dependencies.tsx
```diff
@@ -1,117 +1,77 @@
...
-
-      return <WrapperDependencyDetail key={name} name={name} onLoading={enableLoading} version={version} />;
-    });
+  return (
+    <CardWrap>
+      <CardContent>
+        <Heading variant="subtitle1">{`${title} (${deps.length})`}</Heading>
+        <Tags>
+          {deps.map(([name, version]) => (
+            // eslint-disable-next-line
```
antoinechalifour: This is required because of the `jsx-no-bind` rule which is actually not recommanded (we should discuss it in #168)

src/components/Dependencies/Dependencies.tsx
```diff
@@ -1,117 +1,77 @@
...
+    throw new Error('DetailContext is required');
   }
 
-  private renderDependencies({ packageMeta }): ReactElement<HTMLElement> {
-    const { latest } = packageMeta;
-    const { dependencies, devDependencies, peerDependencies, name } = latest;
+  const { latest } = packageMeta;
+  // FIXME: add dependencies to package meta type
+  // @ts-ignore
+  const { dependencies, devDependencies, peerDependencies, name } = latest;
```
antoinechalifour: We should start working on improving the `PackageMetaInterface` to include all the fields. For now `*dependencies` are not in the type definition

package.json
```diff
@@ -207,5 +205,8 @@
     "type": "opencollective",
     "url": "https://opencollective.com/verdaccio",
     "logo": "https://opencollective.com/verdaccio/logo.txt"
+  },
+  "dependencies": {
+    "react-router-dom": "5.1.2"
```
juanpicado: small reminder, this should be a `devDep`

antoinechalifour: I do not agree. `react-router-dom` Is not a dev tool or a compiler, its code is being used in the compiled, production bundle and thus is a dependency (living with `react` and `react -dom`).

As a rule, anything that is imported by the source file should be a dependency.

yarn.lock
```diff
@@ -10946,23 +10946,23 @@ react-lifecycles-compat@^3.0.4:
...
-    react-router "5.0.1"
+    react-router "5.1.2"
     tiny-invariant "^1.0.2"
     tiny-warning "^1.0.0"
 
-react-router@5.0.1:
-  version "5.0.1"
-  resolved "https://registry.verdaccio.org/react-router/-/react-router-5.0.1.tgz#04ee77df1d1ab6cb8939f9f01ad5702dbadb8b0f"
-  integrity sha512-EM7suCPNKb1NxcTZ2LEOWFtQBQRQXecLxVpdsP4DW4PbbqYWeRiLyV/Tt1SdCrvT2jcyXAXmVTmzvSzrPR63Bg==
+react-router@5.1.2:
```
priscilawebdev: Actually you could remove react-router and use only react-router-dom 😉

src/components/Dependencies/Dependencies.tsx
```diff
@@ -1,117 +1,78 @@
...
-class Dependencies extends Component {
-  public render(): ReactElement<HTMLElement> {
+  if (hasDependencies) {
     return (
-      <DetailContextConsumer>
-        {packageMeta => {
-          return this.renderDependencies(packageMeta as VersionPageConsumerProps);
-        }}
-      </DetailContextConsumer>
+      <Fragment>
```
priscilawebdev:
```suggestion
      <>
```

src/components/Dependencies/Dependencies.tsx
```diff
@@ -1,117 +1,78 @@
...
-                <Heading variant="subtitle1">{`${title} (${deps.length})`}</Heading>
-                <Tags>{this.renderTags(deps, enableLoading)}</Tags>
-              </CardContent>
-            </CardWrap>
-          );
-        }}
-      </DetailContextConsumer>
-    );
+  if (!packageMeta) {
+    throw new Error('DetailContext is required');
```
priscilawebdev: I believe the error message should be : `packageMeta is required`

src/components/Dependencies/Dependencies.tsx
priscilawebdev: sure...there are some rules I also would like to discuss about

src/components/Dependencies/Dependencies.tsx
```diff
@@ -1,117 +1,78 @@
...
+  dependencies: PackageDependencies;
 }
 
-class DepDetail extends Component<DepDetailProps, DepDetailState> {
-  constructor(props: DepDetailProps) {
-    super(props);
-    const { name, version } = this.props;
+const DependencyBlock: React.FC<DependencyBlockProps> = ({ title, dependencies }) => {
+  const { enableLoading } = useContext(DetailContext);
+  const history = useHistory();
```
priscilawebdev: 👏

src/components/Dependencies/Dependencies.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </DetailContextProvider>
+    );
+
+    // Then
+    expect(getByText('verdaccio has no dependencies.')).toBeDefined();
+  });
+
+  test('Renders a link to each dependency', () => {
+    // Given
+    const packageMeta = {
```
priscilawebdev: Maybe you could make the packageMeta reusable?

src/components/Dependencies/Dependencies.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <HashRouter>
+        <DetailContextProvider value={{ packageMeta }}>
+          <Dependencies />
+        </DetailContextProvider>
+      </HashRouter>
+    );
+
+    // Then
+    // FIXME: currently MaterialUI chips do not support the children
+    // prop, therefore it is impossible to use proper links for
```
priscilawebdev: ah? sorry, but I didn't get it...the Chip component accepts an Element as it's child, so you could use `<a>`

package.json
priscilawebdev: I agree with Juan...it should be in the `devDep`. I tried and could not find any react-router-dom in the bundle folder...

antoinechalifour: What do you mean ? `react-router-dom` is used during the application run time so it should definitely stay in the dependencies, not dev dependencies.

Dev dependencies are for tooling around the project, the application router is not tooling, it is a library used by the production code :)

yarn.lock
antoinechalifour: Its already removed :)

src/components/Dependencies/Dependencies.tsx
antoinechalifour: Why are we using `<Fragment>` instead of `<>` ? `<>` is a shortcut from `Fragment` but it is the same :)

src/components/Dependencies/Dependencies.tsx
antoinechalifour: `packageMeta` is only a field of `context` so the message is actually correct :)

src/components/Dependencies/Dependencies.test.tsx
antoinechalifour: I agree ! We should start implementing reusable fixtures for the whole project. My next PR ;)

src/components/Dependencies/Dependencies.tsx
priscilawebdev: It's just a suggestion :-) I prefer `<>`

antoinechalifour: Oh yes sure I thought I wrote the other way ! I'll fix it tomorrow

juanpicado: haha I prefer `<Fragment>`, :P but honestly I started to like `<>`, so +1 for @priscilawebdev

src/components/Dependencies/Dependencies.tsx
juanpicado: What's about `packageMeta is required at DetailContext`? All win.

src/components/Dependencies/Dependencies.test.tsx
juanpicado: Perhaps this helps https://github.com/verdaccio/verdaccio/blob/master/test/unit/__helper/utils.ts

src/components/Dependencies/Dependencies.test.tsx
juanpicado: It's fine for now, thanks for the comment, let's follow up this later on if there is no easy solution now.

src/components/Dependencies/Dependencies.test.tsx
antoinechalifour: How about we use this in a separate PR ?

yarn.lock
priscilawebdev: thank you!

juanpicado:
88% coverage great guys !!

PR #170, opened 2019-10-09
Not Found Component

priscilawebdev:
**Type:** Refactor

**The following has been addressed in the PR:** There is not issue for this PR

**Description:** I had to refactor this component to correct the testing of the Version page. I ended up simplifying the component.


package.json
```diff
@@ -87,8 +87,7 @@
     "react-dom": "16.10.0",
     "react-emotion": "9.2.12",
     "react-hot-loader": "4.12.11",
-    "react-router": "5.0.1",
-    "react-router-dom": "5.0.1",
+    "react-router-dom": "5.1.2",
```
juanpicado: The best is always update dependencies (even if are minor) in separated PR. This helps to identify whether an upgrade has broke something. I learnt the lesson from the past.

src/components/NotFound/NotFound.tsx
```diff
@@ -1,53 +1,44 @@
...
-export type NotFoundProps = RouteComponentProps & WithWidthProps;
+const EmptyPackage = styled('img')({
+  width: '150px',
+  margin: '0 auto',
+});
 
-const HOME_LABEL = 'Home';
+const StyledHeading = styled(Typography)({
+  color: '#4b5e40',
+  marginBottom: 16,
```
juanpicado: I think we have constants for this, right ? Otherwise we should create it.

src/components/NotFound/NotFound.tsx
```diff
@@ -1,53 +1,44 @@
...
-        <Card>{renderList()}</Card>
-      </Inner>
-    </Wrapper>
+    <Box alignItems="center" data-testid="404" display="flex" flexDirection="column" flexGrow={1} justifyContent="center" p={2}>
+      <EmptyPackage alt="404 - Page not found" src={PackageImg} />
+      <StyledHeading className="not-found-text" variant="h4">
+        {NOT_FOUND_TEXT}
+      </StyledHeading>
+      <Button onClick={handleGomHome} variant="contained">
+        {'Go to the home page'}
```
juanpicado: Extract them outside as constants in the same file, so we can use them on testing with react testing library.

src/components/NotFound/Notfound.test.tsx
```diff
@@ -1,18 +1,16 @@
...
 import { BrowserRouter as Router } from 'react-router-dom';
-import { shallow } from 'enzyme';
+import { render } from '@testing-library/react';
 
 import NotFound from './NotFound';
 
-console.error = jest.fn();
-
 describe('<NotFound /> component', () => {
   test('should load the component in default state', () => {
```
juanpicado: Perhaps one `test.todo()` for the onclick, so a next contributor can take care of it :-)

src/components/NotFound/NotFound.tsx
priscilawebdev: theme is coming...let's leave like this now 🙏

src/components/NotFound/NotFound.tsx
priscilawebdev: OK...till we implement i18n 😉

package.json
priscilawebdev: I see...next time then?

PR #171, opened 2019-10-09
Version Page - Replaces class by func.

priscilawebdev:
**Type**: Refactor

I updated the current tests

**Description:**

in order to use react hooks, in this PR I updated the Version Page by changing it from a class to a functional component. I have also splitted it's content so that it is clearer.


priscilawebdev:
Hmmm, I need to check this e2e test. The page is loading here, I just tested everything ... but maybe something doesn't match the test

juanpicado:
@priscilawebdev FYI I added E2E (see workflow details), the page does not load, take your time.

priscilawebdev:
@juanpicado I have only updated the branch and now the e2e tests are passing 🧐 interesting.

src/pages/Version/get-route-package-name.ts
```diff
@@ -0,0 +1,9 @@
+function getRouterPackageName(packageName: string, scope?: string): string {
```
juanpicado: Move this to utilities in a new PR, also router is misleading, it just build a package name based on scope. A `getPackageName` would fit better here.

src/pages/Version/is-package-version-valid.ts
```diff
@@ -0,0 +1,16 @@
+import { PackageMetaInterface } from '../../../types/packageMeta';
+
+function isPackageVersionValid(packageMeta: Partial<PackageMetaInterface>, packageVersion?: string): boolean {
+  if (!packageVersion || typeof packageVersion === 'undefined') {
```
juanpicado: This is also a utilities. The main reason is is much easier to `jest.mock` one file rather 10 in the same unit test.

PR #173, opened 2019-10-10
chore: readded null check

tmkn:
This PR re-adds an explicit `null` test for `isTokenExpire` which got lost in https://github.com/verdaccio/ui/pull/157.


juanpicado:
Thanks @tmkn

PR #174, opened 2019-10-10
Improved Typings

tmkn:
This PR adds explicit type information to places that have been implicitly any.
No actual code changes, just improved typings in various places.

Compiling with `noImplicityAny` is now down to double digits 🎉

### DIFF
--- a/tsconfig.strict.json
+++ b/tsconfig.strict.json
@@ -1,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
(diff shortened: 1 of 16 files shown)


juanpicado:
download

tsconfig.strict.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "noImplicitAny": true
```
juanpicado: ❤️

PR #177, opened 2019-10-11
CircularProgress Component - Introduced ForwardRef

priscilawebdev:
**Type:** Refactor

**The following has been addressed in the PR:** #162

**Unit or Functional tests are included in the PR?**  Not yet

**Description:**

Wrapped Material UI CirularProgress Component with ForwardRef

### DIFF
--- a/src/muiComponents/CircularProgress/CircularProgress.tsx
+++ b/src/muiComponents/CircularProgress/CircularProgress.tsx
@@ -1,7 +1,7 @@
 import React, { forwardRef } from 'react';
 import { default as MaterialUICircularProgress, CircularProgressProps } from '@material-ui/core/CircularProgress';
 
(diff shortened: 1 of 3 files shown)


src/muiComponents/CircularProgress/CircularProgress.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { forwardRef } from 'react';
+import { default as MaterialUICircularProgress, CircularProgressProps } from '@material-ui/core/CircularProgress';
+
+type CircularProgressRef = HTMLElementTagNameMap['div'];
```
tmkn: Why not just use `HTMLDivElement` directly?

priscilawebdev: Yeah, good idea :-) Thanks!

PR #180, opened 2019-10-11
Better type inference for MediaQuery

tmkn:
This PR makes `media.ts` compatible with `noImplicitAny`.

While the code is now a little more verbose, TypeScript is now able to infer the types naturally, without any casts.
Consumption of the `MediaQuery` object stayed the same.


src/utils/styles/media.ts
```diff
@@ -8,26 +8,31 @@ export const breakpoints = {
...
-        @media (${prefix + breakpoints[label] + suffix}) {
-          ${cls};
-        }
-      `;
-    return accumulator;
-  },
-  // eslint-disable-next-line @typescript-eslint/no-object-literal-type-assertion
-  {} as MediaQuery
-);
+function constructMQ(breakpoint: Sizes, args: Interpolation): string {
```
juanpicado: I see too much unnecessary complexity in this method, I'd go for remove it in the future and use media queries by hand, much easy to understand for new contributors.

tmkn: yeah, I just used the existing code and made it more type safe for the compiler, so it doesn't complain as I didn't want to change the usage in the code.

juanpicado: Yes, just was a mental note :-) ... I always wanted to remove it.

PR #188, opened 2019-10-12
Fixed type inference for AppContext

tmkn:
This PR removes a `tsignore` and fixes the underlying complaint by providing better type inference for `AppContext`.

`AppContext.Consumer` can infer the context naturally, however due to the way `AppContext` was typed this was not the case and it complained in some places without `tsignore`. This PR fixes those complaints.


juanpicado:
Thanks !

PR #190, opened 2019-10-12
Added typings for getRecentReleases

tmkn:
This PR changes the signature of `getRecentReleases` from `unknown` to its concrete type.

Adding types slowly but steady :)


src/utils/package.ts
```diff
@@ -4,6 +4,7 @@ import { UpLinks } from '@verdaccio/types';
 import isString from 'lodash/isString';
 import format from 'date-fns/format';
 import distanceInWordsToNow from 'date-fns/distance_in_words_to_now';
+import { Time } from 'types/packageMeta';
```
juanpicado: @tmkn from there this is coming from? 🤔

tmkn: @juanpicado it resolves to
https://github.com/verdaccio/ui/blob/master/types/packageMeta.ts

I just reused the existing import syntax for `types/packageMeta`, it is used in 2 other places like this:
https://github.com/verdaccio/ui/blob/a8deeb9b9d2cb15e61474c904ec4d65f64cc8b1e/src/components/Package/Package.tsx#L5
https://github.com/verdaccio/ui/blob/a8deeb9b9d2cb15e61474c904ec4d65f64cc8b1e/src/utils/calls.ts#L1

I can change it to a relative import?

The reason why the absolute import works is because in the `tsconfig.json` there is a `baseUrl` specified:
https://github.com/verdaccio/ui/blob/a8deeb9b9d2cb15e61474c904ec4d65f64cc8b1e/tsconfig.json#L15
It takes the `baseUrl` and the (absolute) import path to resolve the import. That's why it works.

But as far as I can see we can delete the `baseUrl` option from the `tsconfig.json` as its only purpose is to enable importing `types/packageMeta` via an absolute path and change the imports to a relative one?

juanpicado: I like your approach to be honest, I didn't know :-) but also agree relative is less confusing. Let's change it and delete `baseUrl`.

PR #196, opened 2019-10-13
fix: changes font size for items of the register-info component

### LINKED ISSUE #193: Wrong test on Registry Info
**Describe the bug**

Latest changes seems affected the Typography  of these fields

Screen Shot 2019-10-13 at 9 53 33 AM

Check here previous state: https://registry.verdaccio.org

smithandrewl:
**Type:** bug

**Description:** This changes the font size for items of the register-info component to 1rem. Using the inspector tool, the visible difference between the "correct" and "incorrect" versions of the register-info components was the font size.  The correct/before component had a font-size of 1rem.  This change sets the font size back to 1rem and the text is no longer cut-off at the bottom.

Resolves #193


juanpicado:
Please run `yarn test -u` and upload snapshots.

Issue #199, opened 2019-10-13
Testing: Remove snapshots or limit them to small components rather pages


juanpicado:
**Is your feature request related to a problem? Please describe.**

Avoid

```
    const wrapper = shallow(<UpLinks />);
    expect(wrapper.html()).toMatchSnapshot();
```

It has 0 value with enzyme, is rather annoying to update them.

**Describe the solution you'd like**

- Replace by Testing library, find text or id.
- PR peer component (there are 39 matches in our source code)

anikethsaha:
I think components using styling using `emotion` should not have snapshot testings!
Cause class name using emotion is dynamic and it will change on every run.

juanpicado:
@anikethsaha agree with you, our configuration was not generating random classnames with emotion, so, snapshots should not be affected. If is the current situation, then is a bug.

juanpicado:
Currently seems snapshots are generating random CSS name class, we need to figure out why.

anikethsaha:
> Currently seems snapshots are generating random CSS name class, we need to figure out why.

if this problem persists, maybe migrating to emotion 10 may work as in v10, the styling is inline rather than class name.

juanpicado:
That's the idea, #221  but @sergiohgz has some concerns about the complexity with Typescript.

juanpicado:
This is want I want to avoid ... 

<img width="1630" alt="Screen Shot 2019-10-26 at 5 05 39 PM" src="https://user-images.githubusercontent.com/558752/67621718-e533fc80-f812-11e9-925f-8424702782fd.png">

anikethsaha:
mm... I think this is related to material not particularly with emotion.

anikethsaha:
Also, as using emotion, I saw `jest-emotion` is used to serialize the snapshots, it's being used in the jest config, but as per the docs `expect.addSnapshotSerializer(serializer)` needs to be added in the test files, Is it missed currently or handled in some other way?

agar23:
@juanpicado need any help refactoring the other test's that are not using emotion? I would love to help help out on this front if needed.

juanpicado:
Actually I would like to remove snapshots on pages, just limited to small components and those do not have dates.

PR #206, opened 2019-10-16
chore(tests): remove snapshot test and fix its description (UpLinks.tsx)

hugoazevedosoares:
**Type:** Chore

*  There is a related issue? #199 
*  Unit or Functional tests are included in the PR: Yes

**Description:**
- Remove snapshot test and update it for a more descriptive test.


src/components/UpLinks/UpLinks.test.tsx
```diff
@@ -6,9 +6,9 @@ import { DetailContext } from '../../pages/Version';
 import UpLinks from './UpLinks';
 
 describe('<UpLinks /> component', () => {
-  test('should render the component in default state', () => {
+  test('should return null without packageMeta', () => {
     const wrapper = shallow(<UpLinks />);
-    expect(wrapper.html()).toMatchSnapshot();
+    expect(wrapper.html()).toBeNull();
```
priscilawebdev: Hmmm here you really removed the snapshot...I would rather leave it here, because even if  snapshots don't seem to be efficient, I've already caught many things through them, avoiding possible errors.

hugoazevedosoares: @priscilawebdev I've checked this snapshot before removing it, and it was only a `null` that was been saved. Because of this, I decided to remove it and leave only the assertion, since a `null` snapshot does not bring so much value and it's easier for a developer to ignore it.

priscilawebdev: @hugoazevedosoares it makes sense! I agree :-) thank you!

PR #209, opened 2019-10-17
Converted App component from class to func.

priscilawebdev:
**Type:** Refactor

**Description:**
convert App from Class to Func component.


src/App/App.tsx
```diff
@@ -1,187 +1,105 @@
-import React, { Component, ReactElement } from 'react';
+import React, { useState, useEffect } from 'react';
+import styled from 'react-emotion';
+import Box from '@material-ui/core/Box';
 import isNil from 'lodash/isNil';
-import 'normalize.css';
```
juanpicado: `import 'normalize.css';` we need this

jest/jest.config.js
```diff
@@ -26,5 +26,6 @@ module.exports = {
     'github-markdown-css': '<rootDir>/node_modules/identity-obj-proxy',
     '\\.(png)$': '<rootDir>/node_modules/identity-obj-proxy',
     '\\.(svg)$': '<rootDir>/jest/unit/empty.ts',
+    '\\.(woff|woff2)$': '<rootDir>/jest/unit/fileMock.ts',
```
juanpicado: we don't need this

package.json
```diff
@@ -105,7 +104,6 @@
     "stylelint-processor-styled-components": "1.8.0",
     "stylelint-webpack-plugin": "1.0.3",
     "supertest": "4.0.2",
-    "typeface-roboto": "0.0.75",
```
juanpicado: The fonts are need it

juanpicado:
Codecov is red due several use cases were removed and the current `App.test.tsx` cases are not enough, we need to restore those were removed, user interaction, dialog login, etc etc.

jest/jest.config.js
priscilawebdev: I believe we need this to mock woff and woff2 files, otherwise it won’t pass in the jest test.. I have tried ‘identity-obj-procy’ but it doesn’t really works

package.json
priscilawebdev: They are locally  now 😉

src/App/App.tsx
priscilawebdev: I have created the ResetCss file and I am also using the Baseline file from MUI 🙃 so I believe it is not necessary to have a normalize css dependency

priscilawebdev:
Hi @jotadeveloper thank you for your review. I regarding the App tests..I had to remove them, since I moved many methods and components to smaller components + I have created the AppContext.. because of it many of the tests didn’t make sense anymore..I will check it once again and see if I can create more tests..but I don’t believe so...maybe for the context only

DanielRuf:
Wow, that is a big PR :+1:

Unfortunately I did not yet find the time to take a closer look for a review.

priscilawebdev:
Hi @DanielRuf that is no problem (kein Stress 😉)...I will write also tests for the AppContext, but I am pretty busy this week... I will try to finish this PR over the next weekend..

src/App/App.tsx
```diff
@@ -1,184 +1,106 @@
...
-    scope: window.VERDACCIO_SCOPE || '',
-    showLoginModal: false,
-    isUserLoggedIn: false,
-    packages: [],
-    isLoading: true,
+  /**
+   * Logouts user
+   * Required by: <Header />
+   */
+  const logout = () => {
```
juanpicado: Some recommendations with the functions inside this hook:

- `useCallback`: for `logout`, `loadOnHandler`

It improve performance.

priscilawebdev: True. Thank you!! I will update it then :-)

priscilawebdev:
Thank for the review guys! 🙏I agree with the recommendations and very soon I will open a new PR with these suggested changes 👏😉

PR #219, opened 2019-10-26
refactor:116[PackageList] component is converted to functional

agar23:
**Type:** Refactor

*  There is a related issue? Yes #116 
*  Unit or Functional tests are included in the PR: no

**Description:**

Refactor PackageList component to functional


src/components/PackageList/PackageList.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { Fragment, ReactElement } from 'react';
+import * as React from 'react';
```
juanpicado: We usually uses `import React, { Fragment, ReactElement } from 'react';` format, this is for consistence and using `*` is not a good practice in general.

agar23: @juanpicado  understood I will follow the former pattern for future PRs. Out of curiosity what makes `*` not good practice? is it just cause we are importing everything?

PR #228, opened 2019-10-27
chore: add canary package generator

juanpicado:
**Type:** actions

**Description:**

- Publish a canary version of each PR
- Remove Node 10 and add Node 13 for testing
- Update GH Action main flow, lint before publish


juanpicado:
take a look this @verdaccio/core-team 🚀 

It creates canary versions of each PR, also publish to specific registry and provides a comment with al CLI command just copy and install.

This is an experiment I'm running here in order to move it to Verdaccio, thus is much easier to test.

package.json
```diff
@@ -171,7 +171,6 @@
     "lint:lockfile": "lockfile-lint --path yarn.lock --type yarn --validate-https --allowed-hosts verdaccio npm yarn",
     "coverage:publish": "codecov",
     "pre:webpack": "rimraf static/*",
-    "prepublish": "in-publish && npm run build || not-in-publish",
```
juanpicado: We don't need this. `prepublish` is deprecated anyway.

.github/workflows/canary.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         with:
           node-version: '12.x'
           registry-url: 'https://registry.verdaccio.org'
-      - run: npm publish
+      - run: npm publish -tag canary
```
zkochan: should be --tag

juanpicado: Thanks ! good catch !

.github/workflows/canary.yml
```diff
@@ -0,0 +1,50 @@
+name: CI Canary
+
+on: [pull_request]
+
+jobs:
+  build_test_lint:
+    name: Node Smoke Test Befor Canary
```
priscilawebdev:
```suggestion
    name: Node Smoke Test Before Canary
```

priscilawebdev (approved):

priscilawebdev:
Very good @juanpicado let's move forward 🚀👏

PR #233, opened 2019-10-28
refactor: convert Engine component to hooks

### LINKED ISSUE #116: Using React Hooks (Refactoring, New Components)

bighuggies:
**Type:** refactor

**Description:**

Refactored `<Engine />` to a functional component using hooks, similar to #150 and #156 

Also unrolled the `reduce` function, improving readability.

Snapshots are the same and also tested locally by updating the hardcoded engine data:
image

Resolves #116


src/components/Engines/Engines.tsx
```diff
@@ -1,73 +1,51 @@
...
-      <List subheader={<StyledText variant={'subtitle1'}>{text.split('-').join(' ')}</StyledText>}>
-        <EngineListItem button={true}>
-          {ICONS[text]}
-          <ListItemText primary={heading} />
-        </EngineListItem>
-      </List>
-    );
-  };
-}
+  /* eslint-disable react/jsx-max-depth */
```
bighuggies: Disabled this because I think splitting the render function would harm readability.

Happy to change if anyone disagrees.

src/components/Engines/Engines.tsx
```diff
@@ -1,73 +1,51 @@
...
-      return null;
-    }
+const Engine: React.FC = () => {
+  const { packageMeta } = useContext(DetailContext);
 
-    const engineDict = {
-      'node-JS': engines.node,
-      'NPM-version': engines.npm,
-    };
+  const engines = packageMeta && packageMeta.latest && packageMeta.latest.engines;
```
bighuggies: I saw that TypeScript 3.7 is specified in `package.json`, but `?.` doesn't seem to be supported by the tooling yet :(

src/components/Engines/Engines.test.tsx
```diff
@@ -50,10 +51,9 @@ describe('<Engines /> component', () => {
       latest: {},
     };
 
-    mockPackageMeta.mockImplementation(() => packageMeta);
-
-    const wrapper = mount(<Engine />);
-    expect(wrapper.html()).toEqual('');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
agar23: Why does this need to be `any` ?

src/components/Engines/Engines.test.tsx
```diff
@@ -63,9 +63,8 @@ describe('<Engines /> component', () => {
       },
     };
 
-    mockPackageMeta.mockImplementation(() => packageMeta);
-
-    const wrapper = mount(<Engine />);
-    expect(wrapper.html()).toEqual('');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const wrapper = mount(withEngineComponent(packageMeta as any));
```
agar23: same as above why `any`?

bighuggies: I was aiming for minimal changes to the test fixtures (which were not typed, hence this `any`).

I've updated the fixtures now to make it not necessary.

src/components/Engines/Engines.tsx
```diff
@@ -1,73 +1,51 @@
...
-    );
-  };
-}
+  /* eslint-disable react/jsx-max-depth */
+  return (
+    <Grid container={true}>
+      {engines.node && (
+        <Grid item={true} xs={6}>
+          <List subheader={<StyledText variant={'subtitle1'}>{'node JS'}</StyledText>}>
+            <EngineListItem button={true}>
```
ayusharma: You can remove `button={true}`

src/components/Engines/Engines.tsx
```diff
@@ -1,73 +1,51 @@
...
+              <ListItemText primary={engines.node} />
+            </EngineListItem>
+          </List>
+        </Grid>
+      )}
+
+      {engines.npm && (
+        <Grid item={true} xs={6}>
+          <List subheader={<StyledText variant={'subtitle1'}>{'NPM version'}</StyledText>}>
+            <EngineListItem button={true}>
```
ayusharma: Same here, You can remove button={true}

src/components/Engines/Engines.tsx
```diff
@@ -1,73 +1,51 @@
...
-          <ListItemText primary={heading} />
-        </EngineListItem>
-      </List>
-    );
-  };
-}
+  /* eslint-disable react/jsx-max-depth */
+  return (
+    <Grid container={true}>
+      {engines.node && (
```
ayusharma: Both, `engines.node` and `engines.npm` have same markup. I would suggest to extract it in a component.

src/components/Engines/Engines.test.tsx
```diff
@@ -1,71 +1,60 @@
 import React from 'react';
 import { mount } from 'enzyme';
```
ayusharma: I would suggest to use `@testing-library/react` as some components are already using.  what do you think @juanpicado ?

ayusharma (changes requested):

src/components/Engines/Engines.test.tsx
juanpicado: Yes, agree, `enzyme` is not friendly with hooks yet, so testing-librarty is the way, not mandatory, only if the PR owner agree, it can be a new PR as well.

src/components/Engines/Engines.tsx
juanpicado: This should be enforced by `eslint`, I'd suggest let it go and add the plugin or rule and change all places at a time.

src/components/Engines/Engines.tsx
juanpicado: That's odd, I did not tried yet. I'll double check.

src/components/Engines/Engines.tsx
juanpicado: Agree, rules has exceptions, this one is one of them.

src/components/Engines/Engines.test.tsx
bighuggies: When doing a refactor I prefer to keep the tests the same as much as possible, changing the tests at the same time invalidates their purpose (guarding against regressions).

I definitely prefer RTL, will be happy to open a PR for that once this is merged

src/components/Engines/Engines.tsx
bighuggies: I think it was prettier that was giving me issues, 1.19 with support for 3.7 isn't released yet: https://github.com/prettier/prettier/blob/master/CHANGELOG.unreleased.md

src/components/Engines/Engines.tsx
bighuggies: I think @ayusharma is referring to the fact that the `EngineListItem` styled component doesn't accept a `button` prop.

I left it because removing it breaks the snapshot tests, which I consider bad practice when doing a refactor like this.

src/components/Engines/Engines.tsx
bighuggies: I decided against it because I think the extra indirection adds more cognitive load than the benefit the abstraction gives.

src/components/Engines/Engines.tsx
priscilawebdev: Really? 😞 I didn't test the new features yet, but I thought we could already use it ... so let's update prettier :D

src/components/Engines/Engines.tsx
priscilawebdev: Hmmm that is a good argument ... but if we keep write the same code like this, example we will have more engines, I would extract it in a new component as @ayusharma suggested... for now is fine.

src/components/Engines/Engines.test.tsx
```diff
@@ -1,71 +1,60 @@
...
+import { DetailContext } from '../../pages/Version';
+import { PackageMetaInterface } from '../../../types/packageMeta';
+
 import Engine from './Engines';
 
 jest.mock('./img/node.png', () => '');
 jest.mock('../Install/img/npm.svg', () => '');
 
-const mockPackageMeta: jest.Mock = jest.fn(() => ({
+const withEngineComponent = (packageMeta: PackageMetaInterface): JSX.Element => (
```
priscilawebdev: I am not sure I like the name "withEngineComponent", because that is not really a HOC...it's just a func with a param..do you know what I mean?..

priscilawebdev (commented):
@bighuggies thanks for contributing :-) I really like that you added a print of your changes.. I will do the same next time. Having something visual helps during review

src/components/Engines/Engines.tsx
priscilawebdev: Hi all! I just checked and the ListItem component from Material UI does accepted the prop button..we should change it if we don't want to use the ButtonBase only : 
image

So IMO it's fine if you leave like it is now...we will check all cases in our design system

priscilawebdev: https://material-ui.com/api/list-item/

priscilawebdev (dismissed):

src/components/Engines/Engines.test.tsx
bighuggies: Good point! I didn't notice the conceptual conflict with HOCs, I was being consistent with the suggestion from my other PR: https://github.com/verdaccio/ui/pull/150#discussion_r331141170

In this case I think the abstraction is a bit pointless since the tests and fixtures are so short, I've inlined the JSX into each test and it seems readable. This also avoids the hard naming problem 😅

src/components/Engines/Engines.tsx
priscilawebdev: regarding this `lint` rule, I believe we should discuss about it. I find this rule very annoying ... I would also prefer to write `button` only

src/components/Engines/Engines.test.tsx
priscilawebdev: @bighuggies thank you!

src/components/Engines/Engines.test.tsx
```diff
@@ -1,71 +1,66 @@
...
+import { DetailContext } from '../../pages/Version';
+import { PackageMetaInterface } from '../../../types/packageMeta';
+
 import Engine from './Engines';
 
 jest.mock('./img/node.png', () => '');
 jest.mock('../Install/img/npm.svg', () => '');
 
-const mockPackageMeta: jest.Mock = jest.fn(() => ({
+const mockPackageMeta = (): PackageMetaInterface => ({
```
priscilawebdev: IMO this should be an object, not a func.

src/components/Engines/Engines.test.tsx
```diff
@@ -1,71 +1,66 @@
...
   });
 
   test('should render the component when there is no engine key in package meta', () => {
-    const packageMeta = {
-      latest: {},
-    };
+    const packageMeta = mockPackageMeta();
 
-    mockPackageMeta.mockImplementation(() => packageMeta);
+    delete packageMeta.latest.engines;
```
priscilawebdev: Hmm...here you are mutating the object.. something I wouldn't do... I would filter it

juanpicado: @bighuggies you can try with `Object.assign` or spread the props, so you don't have to mutate the object.

```
const packageMeta = Object.assign({}, mockPackageMeta(), {
  latest: { engines: undefined })
};
```
This is more verbose, but, according the standard we promote here, no mutations.

src/components/Engines/Engines.test.tsx
juanpicado: The code was already there, in this case, we can ignore this part for now.

bighuggies: Using a function ensures the object is new in each test, preventing accidental contamination caused by mutations.

src/components/Engines/Engines.test.tsx
bighuggies: Because it's a deeply nested object, the non-mutational form would have to be a bit more complicated:

```typescript
    const initialPackageMeta = mockPackageMeta();
    const packageMeta = {
      ...initialPackageMeta,
      ...{
        latest: {
          ...initialPackageMeta.latest,
          engines: undefined,
        },
      },
    };
```

I think the mutational form is easier to read and conveys the intention of the test much more clearly.

Mutations are bad practice in general but I think tests like this are an exception. They run in isolation and the point is to be readable.

Regardless, I've changed `mockPackageMeta` to take the `engine` object as a param, I think it's slightly less clear but avoids mutations.

src/components/Engines/Engines.test.tsx
bighuggies: I've changed this now - the function takes a param to set up the `engine` prop in the test data

juanpicado (approved):
Great !!! 🚀 👻 👻 🎃 🎃

PR #134, opened 2019-09-28
fix: correctly load font files - closes #128

DanielRuf:
**Type:**
bug

**Description:**
This should fix the wrong paths for the Roboto fonts.

Resolves #128

PR #240, opened 2019-11-07
Refactor: Created Reset CSS and added local fonts

priscilawebdev:
**Type:** Refactor and Fix

**Description:**
 I noticed an error in the console regarding to a font load. I have fixed it.

I also replaced two dependencies(typeface-roboto and normalize.css) with locally fonts and reset css files :) . I believe that we don't really extra dependencies for these simple things.

Please let me know what do you think. 😉


package.json
```diff
@@ -166,7 +164,7 @@
     "test": "cross-env NODE_ENV=test BABEL_ENV=test TZ=UTC jest --config ./jest/jest.config.js --maxWorkers 2 --passWithNoTests",
     "test:size": "bundlesize",
     "lint": "npm run lint:js && npm run lint:css && npm run lint:lockfile",
-    "lint:js": "npm run type-check && eslint . --ext .js,.ts,.tsx",
+    "lint:js": "npm run type-check && eslint . --ext .js,.ts,.tsx --fix",
```
juanpicado: `--fix` I guess you forgot this

priscilawebdev: no I didn't :-) Actually as I am having problems with the actual version of eslint and pretty...I am using this command quite often and I thought it would be helpful for another devs..especially before pushing

juanpicado: That's actually the job of `husky` and `lint-staged`, for every filed changed `prettier` pass through each file.

```
 "lint-staged": {
    "relative": true,
    "linters": {
      "*.{js,tsx,ts}": [
        "eslint . --ext .js,.ts,.tsx",
        "prettier --write"
      ],
      "*": [
        "detect-secrets-launcher --baseline .secrets-baseline",
        "git add"
      ]
    },
    "ignore": [
      "*.json"
    ]
  },
````

The script `lint:js` is being executed in the CI, this might cause unexpected errors due the files are modified in such environment. We don't need the `fix` flag.

Comment (priscilawebdev) [reply package.json]: ah yes..it's true! thank you!

**Comment by juanpicado** [comment]
I'm still not agree removing `"normalize.css": "8.0.1",` and `"typeface-roboto": "0.0.75",`, frankly I don't see the benefit as we discussed previously.

**Comment by juanpicado** [comment]
Unfortunately this not have my approval, previously we discussed that there reset **must** comes from dependencies and still are here
https://github.com/verdaccio/ui/pull/240/files#diff-684c724a43a7ac26119a730764a995eeR7

Now we have reset CSS duplicated

```
import 'normalize.css';
import 'typeface-roboto/index.css';

import ResetCSS from './ResetStyles';
``

Please rollback those changes.

juanpicado:
One more thing, `import 'typeface-roboto/index.css';` this is not fully well, if the `typeface` team changes the `main` endpoint in their packages, the url won't work, the best is rely on the `main` field in the `package.json`, thus `import 'typeface-roboto` is the best option here.

DanielRuf:
Afaik there will not be such breaking changes in there. Let's do another PR to resolve this.

And I remember that I had some issue with jest, webpack and so on when I did not specify the `.css` part.

See https://github.com/KyleAMathews/typefaces/search?q=import+css&type=Issues and especially https://github.com/KyleAMathews/typefaces/issues/113

https://github.com/verdaccio/ui/pull/134 ;-)

DanielRuf:
> Please rollback those changes

Let's open a new PR for this =)

priscilawebdev:
@juanpicado there are some css reset things I would like to apply that doesn't come with the `normalize.css`. Example the font... if you check the current registry.veerdaccio.org and inspect some h2 you are going to see that they are not `Roboto`. That is the reason I kept the `ResetStyles` component. Regarding the `typeface-roboto` I just kept it as it was, although I agree with you..

DanielRuf:
There is a good reason for the `.css` import (because of jest and so on).

See my last comment.

juanpicado:
> @juanpicado there are some css reset things I would like to apply that doesn't come with the `normalize.css`. Example the font... if you check the current registry.verdaccio.org and inspect some h2 you are going to see that they are not `Roboto`. That is the reason I kept the `ResetStyles` component. Regarding the `typeface-roboto` I just kept it as it was, although I agree with you..

I see your point, but I think we have avoid the miss understanding with a better description in your PR about what you were trying to achieve. I'd suggest you rename `resetStyles` to something more meaningful, as `applyFontStyles` or something that describe what really does, because reset styles means reset default styles of the browser.

Anyway, I run some test on master and all is ok.

PR #300, opened 2019-11-21
fix: correctly reference registry url from options

pmmmwh:
**Type: Bug Fix**

**Description:**

Currently, the repository URL is not being correctly referenced in the Registry Info dialog if the web UI is not loaded at root. This is discovered and illustrated [here](https://github.com/n4bb12/verdaccio-github-oauth-ui/pull/38#issuecomment-556524702). This PR uses `window.__VERDACCIO_UI_OPTIONS` to correctly get the base url so that loading will not affect the display of info.


juanpicado:
Thank u so much for this !

anikethsaha (dismissed):
Looks good 👍

juanpicado:
Only test are failing

anikethsaha:
I think tests need to be updated.

image

registry URL should not contain the path, right?

- `http://localhost/-/web/detail` is not a registry URL.
- `http://localhost` it is 

cc @juanpicado , what do you think ?

src/utils/url.ts
```diff
@@ -15,8 +15,7 @@ export function isEmail(email: string): boolean {
 }
 
 export function getRegistryURL(): string {
-  // Don't add slash if it's not a sub directory
-  return `${location.origin}${location.pathname === '/' ? '' : location.pathname}`;
+  return (window as any).__VERDACCIO_BASENAME_UI_OPTIONS.base;
```
priscilawebdev:
```suggestion
  return window.__VERDACCIO_BASENAME_UI_OPTIONS.base;
```

you don't need to cast window as any here, since there is already a type __VERDACCIO_BASENAME_UI_OPTIONS for window in the file types/index.ts. Please see: 
image

priscilawebdev (changes requested):

priscilawebdev:
thanks for your contribution @pmmmwh  :) 

Unfortunately tests are failing..

src/utils/url.ts
pmmmwh: Okies! I will update that.

pmmmwh:
> thanks for your contribution @pmmmwh :)
> 
> Unfortunately tests are failing..

I'll check and update accordingly - I think it is mostly because of `window.__VERDACCIO_BASENAME_UI_OPTIONS` not containing the correct information.

hugoazevedosoares:
Looks good to me

priscilawebdev:
Now it looks good! Thank you @pmmmwh

juanpicado:
Sorry guys, but, the info is broken.

Screen Shot 2019-11-24 at 7 52 52 PM

I'd suggest a new PR with :

* E2E that verify this works
* Please thing in all use cases

```
export function getRegistryURL(): string {
  return window.__VERDACCIO_BASENAME_UI_OPTIONS.base;
}
```
We don't use server rendering, thus, the base only include that, the base domain, but does not include the **path** or possible queries. As consequence the `window.__VERDACCIO_BASENAME_UI_OPTIONS.base` will return always `/` as you can see in the screenshot above.

priscilawebdev:
Hi @juanpicado that is true..I have just checked and it's broken :( I've revert the changes

juanpicado:
It's fine, I think the problem is bigger that is seems so, it might need some changes in the backend. I introduced this bug with this https://github.com/verdaccio/verdaccio/pull/1457

pmmmwh:
> Sorry guys, but, the info dialog is broken. #310
> 
> Screen Shot 2019-11-24 at 7 52 52 PM
> 
> I'd suggest a new PR with :
> 
> * E2E that verify this works
> * Please think in all use cases and scenarios
> * Verify this works with `url_prefix` prop https://verdaccio.org/docs/en/configuration#url-prefix
> 
> ```
> export function getRegistryURL(): string {
>   return window.__VERDACCIO_BASENAME_UI_OPTIONS.base;
> }
> ```
> 
> We don't use server rendering, thus, the base only include that, the base domain, but does not include the **path** or possible queries. As consequence the `window.__VERDACCIO_BASENAME_UI_OPTIONS.base` will return always `/` as you can see in the screenshot above.

Oops - I'm sorry about that. I'll see if there's another way to achieve the correct behaviour.

PR #306, opened 2019-11-23
chore: enable optional chaining and nullish

juanpicado:
It enables 

- https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-optional-chaining
- https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-nullish-coalescing-operator

Future updates of `@verdaccio/babel-preset` should include this by default.

⚠️ Ref https://github.com/verdaccio/monorepo/issues/274

**Type:**

*  There is a related issue?

**Description:**


src/components/Engines/Engines.tsx
```diff
@@ -13,7 +13,7 @@ import node from './img/node.png';
 const Engine: React.FC = () => {
   const { packageMeta } = useContext(DetailContext);
 
-  const engines = packageMeta && packageMeta.latest && packageMeta.latest.engines;
+  const engines = packageMeta?.latest?.engines;
```
juanpicado: PoC that the operator works.

anikethsaha:
Wow, looking forward to this 👍 
Is there any external linter plugins/config/rule required to support this?

juanpicado:
In Visual Code is displayed like this, I ignore the reason, while in ...
Screen Shot 2019-11-23 at 6 34 22 PM

Webstorm looks fine.
Screen Shot 2019-11-23 at 6 34 29 PM

sergiohgz:
See which version of TS is using Vscode (statusBar, right zone). I remember that, by default, it will use 3.6.x, but I'm not sure

tmkn:
It should use the installed version of TypeScript instead of the built in one from VSCode, I added the setting for it: https://github.com/verdaccio/ui/blob/58cf730b983425d20d11eb7cb878beaf1f325a16/.vscode/settings.json#L9
:confused:

anikethsaha:
What I think is as we are using `typescript-eslint` parser, the rules are not releases yet I guess.
Parsing is supported but not the rules.
https://github.com/typescript-eslint/typescript-eslint/pull/1253


Tracking issues 
https://github.com/typescript-eslint/typescript-eslint/issues/1051
https://github.com/typescript-eslint/typescript-eslint/issues/1052

I think  Babel-eslint parser does support the rules as well.

src/components/Engines/Engines.tsx
```diff
@@ -13,7 +13,7 @@ import node from './img/node.png';
...
 
   if (!engines || (!engines.node && !engines.npm)) {
```
tmkn:
```suggestion
  if (!engines?.node && !engines?.npm) {
```
I know it's only a PoC but since it's right below, this `if` can also be simplified with optional chaining

tmkn:
```suggestion
  if (!engines?.node && !engines?.npm) {
```
I know it's only a PoC but since it's right below, this `if` can also be simplified with optional chaining

priscilawebdev:
very  nice!! :-) Yeah I believe VsCode doesn't support it yet, but they are working on it.. I will check the status ;-)

priscilawebdev:
I checked...please see their last updated (October 2019) 
image

.yarnrc
```diff
@@ -1 +1,2 @@
 save-prefix ""
+registry "https://registry.verdaccio.org"
```
priscilawebdev: 👏

priscilawebdev:
After installing the dependency. it's looks fine:
image

sergiohgz:
> What I think is as we are using `typescript-eslint` parser, the rules are not releases yet I guess.
> Parsing is supported but not the rules.
> [typescript-eslint/typescript-eslint#1253](https://github.com/typescript-eslint/typescript-eslint/pull/1253)
> 
> Tracking issues
> [typescript-eslint/typescript-eslint#1051](https://github.com/typescript-eslint/typescript-eslint/issues/1051)
> [typescript-eslint/typescript-eslint#1052](https://github.com/typescript-eslint/typescript-eslint/issues/1052)
> 
> I think Babel-eslint parser does support the rules as well.

Qe should not be using Babel Eslint parser, as it doesn't support well TS rules. For that, typescript-eslint parser is the solution

anikethsaha:
> Qe should not be using Babel Eslint parser, as it doesn't support well TS rules. For that, typescript-eslint parser is the solution

yeah, 
I guess we can wait for  typescript-eslint to complete the rules and then we can do the required changes in the eslint config if required 👍

juanpicado:
Thanks guys for your feedback, we will update those things :)

PR #307, opened 2019-11-23
Improvement/types

tmkn:
**Description:**
This PR improves once again the typings.
`strict` mode errors are down from 91 to 50 🚀


src/components/Login/Login.tsx
```diff
@@ -131,22 +130,19 @@ export default class LoginModal extends Component<Partial<LoginModalProps>, Logi
 
   public submitCredentials = async () => {
     const { form } = this.state;
-    const username = (form.username && form.username.value) || '';
-    const password = (form.password && form.password.value) || '';
+    const username = form.username?.value ?? '';
```
tmkn: null coalescing operator and optional chaining at work 💪

src/components/Login/Login.tsx
```diff
@@ -98,31 +100,28 @@ export default class LoginModal extends Component<Partial<LoginModalProps>, Logi
...
       {
-        form: Object.keys(form).reduce(
-          (acc, key) => ({
-            ...acc,
-            ...{ [key]: { ...form[key], pristine: false } },
-          }),
-          { username: {}, password: {} }
-        ),
+        form: {
+          username: { ...form.username, pristine: false },
```
tmkn: I opted for a simple assignment since the keys weren't dynamic and the reduce construct was hard to understand for TypeScript.

src/components/Login/Login.tsx
```diff
@@ -131,22 +130,19 @@ export default class LoginModal extends Component<Partial<LoginModalProps>, Logi
 
   public submitCredentials = async () => {
     const { form } = this.state;
-    const username = (form.username && form.username.value) || '';
-    const password = (form.password && form.password.value) || '';
+    const username = form.username?.value ?? '';
+    const password = form.password?.value ?? '';
     const { onSubmit } = this.props;
     if (onSubmit) {
       await onSubmit(username, password);
```
tmkn: While not an error, vscode suggested to remove the unnecessary `await`.
Turns out `onSubmit` was typed with `void` and not `Promise<void>` for its return value, fixed that as well
https://github.com/verdaccio/ui/blob/00d6223ad85f753c31e21e8eb3e85120d665f38a/src/components/Login/Login.tsx#L33

src/components/Login/Login.tsx
juanpicado: nice !! glad you can use it already !

src/components/Login/Login.tsx
priscilawebdev: well done.

src/components/RegistryInfoContent/RegistryInfoContent.tsx
```diff
@@ -66,7 +66,7 @@ const RegistryInfoContent: React.FC<Props> = props => {
   };
 
   /* eslint react/prop-types:0 */
-  const TabContainer = ({ children }): JSX.Element => {
+  const TabContainer: React.FC = ({ children }) => {
```
priscilawebdev: 👏

src/utils/test-enzyme.ts
```diff
@@ -4,13 +4,13 @@ import ThemeProvider from '../design-tokens/ThemeProvider';
...
 
-const shallowWithTheme = (element: React.ReactElement<any>, ...props): any =>
+const shallowWithTheme = (element: React.ReactElement<any>, ...props: undefined[]): any =>
   shallow(element, {
     wrappingComponent: ThemeProvider,
     ...props,
   });
 
-const mountWithTheme = (element: React.ReactElement<any>, ...props): any =>
+const mountWithTheme = (element: React.ReactElement<any>, ...props: undefined[]): any =>
```
priscilawebdev: Sorry, but I am not sure I agree with this type 🧐

src/components/Search/Search.tsx
```diff
@@ -25,7 +25,7 @@ export type handleClickSearch = (
   event: KeyboardEvent<HTMLInputElement>,
   { suggestionValue, method }: { suggestionValue: object[]; method: string }
 ) => void;
-export type handleFetchPackages = ({ value: string }) => Promise<void>;
+export type handleFetchPackages = (arg: { value: string }) => Promise<void>;
```
priscilawebdev: Please, why do you find "arg:" necessary? I believe to be irrelevant

src/components/Search/Search.test.tsx
```diff
@@ -246,7 +246,7 @@ describe('<Search /> component test', () => {
       const getDetailPageURL = jest.fn(() => 'detail/page/url');
       jest.doMock(URL_FILE_PATH, () => ({ getDetailPageURL }));
 
-      const suggestionValue = [];
+      const suggestionValue: unknown[] = [];
```
priscilawebdev: I would prefer to create an interface for the suggestionValue, but since I am refactoring the AutoComplete component we can leave it like that

src/components/Search/Search.test.tsx
```diff
@@ -155,7 +155,7 @@ describe('<Search /> component test', () => {
     beforeEach(() => {
       jest.resetModules();
       jest.doMock('lodash/debounce', () => {
-        return function debounceMock(fn) {
+        return function debounceMock(fn: unknown) {
```
priscilawebdev: Please are you sure about this type?

src/components/Repository/Repository.test.tsx
```diff
@@ -19,7 +20,7 @@ const mockPackageMeta: jest.Mock = jest.fn(() => ({
 }));
 
 jest.mock('../../pages/Version', () => ({
-  DetailContextConsumer: component => {
+  DetailContextConsumer: (component: { children: (arg0: { packageMeta: PackageMetaInterface }) => void }) => {
```
priscilawebdev: sorry, but I am not so sure I like it...maybe there is a simpler way?

src/components/Search/Search.tsx
tmkn: Haha this one was tricky, I didn't encounter this before 😉 
but that particular syntax didn't work as intended.
tldr: It only **renamed** `value` to `string`.
It didn't say that `value` is of type `string`.
see https://github.com/microsoft/TypeScript/issues/9657

To interpret `{ value: string }` as a type like intended and not a rename in a destructure assignment (which is apparently possible) you need to add the argument name.

src/components/Search/Search.test.tsx
tmkn: Since this is only a type to satisfy the unit test and called internally by jest aka never called manually by production code I just used the fastest route to satisfy the compiler.

src/components/Repository/Repository.test.tsx
tmkn: It was auto generated by vscode, it can now generate types based on the usage 🚀 
Didn't think it would work but it worked.
I can properly type this if wanted.

src/utils/test-enzyme.ts
tmkn: Again since this is only used to type a unit test and not called by production code I took the fastest route to satisfy the compiler.

priscilawebdev: Hm..I understand your point. I would do so

image

....but we have this lint rule

The thing is that if the type is "any" or the build is not satisfied, then we know that we need to right the correct type, but when we introduce, for instance undefined[], we might not see where we should fix in the future... do you understand? so even if it's not production code, I would appreciate to have it correct..otherwise leave as it is

src/components/Search/Search.tsx
priscilawebdev: Very interesting! but in this case i would then do: `({value}: { value: string})`. It's more readable IMO, otherwise people can get confused

src/components/Search/Search.test.tsx
priscilawebdev: I see...but maybe it's better to right a function type, since this is a function right? for example: `fn: () => void`. some investigation might be need to check what we could write to replace void..

priscilawebdev: Maybe just a `Function` type would be also fine

src/components/Repository/Repository.test.tsx
priscilawebdev: Hm... I refactored this component today... I've removed this part of the code, so that is fine :-)

src/components/Login/Login.tsx
DanielRuf: Do we transpile this? It is only availble in Node 13+ afaik.

src/components/Repository/Repository.test.tsx
DanielRuf: Well but then we have to resolve the conflicts now ;-)

priscilawebdev:
@tmkn please what is the status of this PR?

tmkn:
@priscilawebdev will update this pr this weekend, just don't know whether its worth the effort to merge and resolve the conflicts or start with a fresh master since the type improvements where straight forward.

juanpicado:
As you wish, I would do be same

tmkn:
Will open a new PR for the improved typings so that it doesn't drag along all the accumulated baggage from this.

priscilawebdev:
I would do the same. Thanks @tmkn :)

PR #308, opened 2019-11-24
fix: formatDate

tmkn:
**Description:**
On Windows the tests failed for `formatDate`. It somehow took the timezone offset into account.
Adding the timezone offset before passing it to `date-fns/format` fixes it for Windows while remaining intact for Mac.

Since the tests now fully work on Windows we can add them to the CI pipeline?


tmkn:
Also:
https://github.com/verdaccio/ui/blob/b2ea2e416740ff33f0cd6533bf970dc5a34ebc71/src/utils/package.ts#L42
is only used in unit tests. There are 5 tests for it but it's not used anywhere else.
If it doesn't serve a function it can be removed.

juanpicado:
@tmkn good catch, Mmmmm .... if `formatDate` I would like to know when was removed ... if is really not need it, we should remove it

tmkn:
@juanpicado `formatDate` is still used, `formatRepository` is only used in unit tests.

juanpicado:
sorry the delay @tmkn I think I screw this PR :(

tmkn:
@juanpicado no worry, should I remove `formatRepository` in the next PR? Did you take a look?

PR #339, opened 2019-12-03
Search Component - Replaced class by func. comp

priscilawebdev:
**Type:** Refactor

**Description:**

in order to use react hooks, in this PR I updated the Search component by changing it from a class to a functional component.

I have also turned off this lint rule: `@typescript-eslint/explicit-function-return-type`, because 
it is not really helpful. please do you agree with me?


src/components/Search/Search.test.tsx
```diff
@@ -1,272 +1,147 @@
...
+import { render, fireEvent, waitForElement } from '../../utils/test-react-testing-library';
+import api from '../../utils/api';
 
 import Search from './Search';
 
-const SEARCH_FILE_PATH = './Search';
-const API_FILE_PATH = '../../utils/calls';
-const URL_FILE_PATH = '../../utils/url';
+/* eslint-disable verdaccio/jsx-spread */
+const ComponentToBeRendered: React.FC = () => (
```
juanpicado: `ComponentToBeRendered` :D ... I hope ..

src/components/Search/Search.tsx
```diff
@@ -1,139 +1,91 @@
...
-        suggestions={suggestions}
-        suggestionsError={error}
-        suggestionsLoaded={loaded}
-        suggestionsLoading={loading}
-        value={search}
-      />
-    );
-  }
+const Search: React.FC<RouteComponentProps> = ({ history }) => {
+  const [suggestions, setSuggestions] = React.useState([]);
```
juanpicado: No blocking, but in the other places we use the `useState` from the import.

src/components/Search/Search.tsx
```diff
@@ -1,139 +1,91 @@
...
+  const [loading, setLoading] = React.useState(false);
+  const [requestList, setRequestList] = React.useState<Array<{ abort: () => void }>>([]);
 
   /**
    * Cancel all the requests which are in pending state.
    */
-  private cancelAllSearchRequests: cancelAllSearchRequests = () => {
-    this.requestList.forEach(request => request.abort());
-    this.requestList = [];
+  const cancelAllSearchRequests = () => {
```
juanpicado: `useCallback` if you consider works here

juanpicado: Also below

priscilawebdev: Thank you for the suggestion :-) I checked, but  unfortunately in this case the callback doesn't help, it just adds more complexity

PR #345, opened 2019-12-06
Replaced date-fns with dayjs

priscilawebdev:
**Type:** Refactor 

**Description:**

I noticed that we have many rooms to improve on performance, accessibility, etc. I will open an issue for it, but today I have started replacing the date.fn library with dayjs. 

Why? 

Comparing the  size of the two libraries, the difference is very significant. Please see: 

image
image

...and although we don't use this library much, it has a considerable size in our bundle. 

As you can see while [date-fn](https://github.com/date-fns/date-fns) has 16.8  kB  Minified + Gzipped , [dayjs](https://github.com/iamkun/dayjs) has only 2.8 kB  Minified + Gzipped.

DayJs also has all the methods we use 👏😀 In this PR, I have replaced the following methods:

addHours(new Date(), limit).getTime() ->  dayjs(new Date()).add(amount, 'hour').valueOf()
formatDistanceToNow(new Date(lastUpdate)) -> dayjs(new Date(lastUpdate)).fromNow()
format(new Date(lastUpdate), TIMEFORMAT) -> dayjs(new Date(lastUpdate)).fromNow()

Please let me know what you think about :)

ps: the library is also in active development 😉


src/utils/package.test.ts
```diff
@@ -71,7 +71,7 @@ describe('formatDateDistance', (): void => {
     const date2 = dateTwoMonthsAgo();
     // FIXME: we need to review this expect, fails every x time.
     // expect(formatDateDistance(date1)).toEqual('about 2 months');
-    expect(formatDateDistance(date2)).toEqual('2 months');
+    expect(formatDateDistance(date2)).toEqual('2 months ago');
```
juanpicado: does `formatDateDistance` add the `ago` by default? 😮

src/utils/package.ts
```diff
@@ -52,11 +55,11 @@ export function formatRepository(repository: any): string | null {
 }
 
 export function formatDate(lastUpdate: string | number): string {
-  return format(new Date(lastUpdate), TIMEFORMAT);
+  return dayjs(new Date(lastUpdate)).format('DD.MM.YYYY, HH:mm:ss');
```
juanpicado: You can reuse `TIMEFORMAT` here, it is above 👍

priscilawebdev: Oh True! I forgot 🙂

src/utils/package.test.ts
priscilawebdev: Yes :D

PR #360, opened 2019-12-11
Refactor: Updated developers component structure

priscilawebdev:
**Type:** Fix 

**Description:**

the developers array had a duplicate e-mail values. In this PR I made it unique so then we won't see the error related to the unique key in the loop anymore.


src/components/Developers/Developers.tsx
```diff
@@ -1,60 +1,85 @@
...
-  return renderDevelopers(developerList, packageMeta);
+  const developers = useMemo(() => getUniqueDeveloperValues(detailContext.packageMeta?.latest[type]), [
+    detailContext.packageMeta,
+    type,
+  ]);
+  const [visibleDevelopersMax, setVisibleDevelopersMax] = useState(visibleMax);
+  const [visibleDevelopers, setVisibleDevelopers] = useState(developers);
+
+  useEffect(() => {
+    if (!developers) return;
```
juanpicado: `eslint` should enforce usage of `{` `}` ,

src/components/Developers/Developers.tsx
```diff
@@ -1,60 +1,85 @@
...
+    <>
+      <StyledText variant={'subtitle1'}>{type}</StyledText>
+      <StyledBox display="flex" flexWrap="wrap" margin="10px 0 10px 0">
+        {visibleDevelopers.map(visibleDeveloper => (
+          <Tooltip key={visibleDeveloper.email} title={visibleDeveloper.name}>
+            <Avatar alt={visibleDeveloper.name} src={visibleDeveloper.avatar} />
+          </Tooltip>
+        ))}
+        {visibleDevelopersMax < developers.length && (
+          <Fab onClick={() => setVisibleDevelopersMax(visibleDevelopersMax + VISIBLE_MAX)} size="small">
```
juanpicado: You can extract this to a function above and wrap it with `useCallback` and avoid create a closure here `() => setVisibleDevelopersMax(visibleDevelopersMax + VISIBLE_MAX)`

src/components/Developers/Developers.tsx
priscilawebdev: I will check it and if necessary open a new PR with the rule 😉

Issue #361, opened 2019-12-12
Display funding on UI


juanpicado:
**Is your feature request related to a problem? Please describe.**

Currently the metadata has a new field, funding, supported by the command `npm fund` and I want to display it in the UI.

```
 "10.0.5": {
            "name": "preact", 
            "amdName": "preact", 
            "version": "10.0.5",          
            "license": "MIT", 
            "funding": {
                "type": "opencollective", 
                "url": "https://opencollective.com/preact"
            }, 
```

Context: https://github.com/npm/rfcs/pull/54

This also requires, update https://github.com/verdaccio/monorepo/tree/master/core/types

**Describe the solution you'd like**

1. Just below the package name
2. Add as a button ( but it has less visibility)
3. Below installation panel
4. Propose something else 🤔 

![Verdaccio_-__verdaccio_babel-preset](https://user-images.githubusercontent.com/558752/70687482-9e5d7180-1caf-11ea-8c5a-3d0ab6ddefe0.png)

**Additional context**

I want Verdaccio contribute to funding packages across al interfaces.
**On click a new tab must be open**

priscilawebdev:
I really like the idea!! Can I take it over?  @verdaccio/core-team  😀

DanielRuf:
Great idea.
This would be cool @priscilawebdev =)

hugoazevedosoares:
Nice one. I think the number 2 proposal will look better, but on proposal 3 it may be more visible.

DanielRuf:
I think proposal 3 is best (clearly shown as external link).

Wasn't there also a demo of npmjs with the funding links?

anikethsaha:
+1 for this.
Lets go with 3rd with a wide button like npm is having it now.

juanpicado:
Thanks guys I think proposal 3 would be.

priscilawebdev:
I still need to update the funding use cases, but as the funds button is already there, I think I can close this issue ☺️

juanpicado:
The PR #375  only add a button with no behavior as far I can see, so I reopen this until is fully working.

ref https://github.com/verdaccio/ui/pull/375

DanielRuf:
Isn't it already working there with the sryled link? https://github.com/verdaccio/ui/pull/375/files (I see more changes again than just the addition of the button, we should do refactorings in separate PRs).

What is currently missing?

juanpicado:
[...]

PR #375, opened 2019-12-17
Feat: added "Fund this package" button

priscilawebdev:
**Type:** Feat

The following has been addressed in the PR: #361

**Description:** Introduced "Fund this package" button. 

image


juanpicado:
@priscilawebdev it has conflicts ..

src/components/DetailSidebar/DetailSidebarFundButton.tsx
```diff
@@ -0,0 +1,46 @@
...
+  marginRight: 3,
+});
+
+interface Props {
+  to: string;
+}
+
+/* eslint-disable react/jsx-no-bind */
+const DetailSidebarFundButton: React.FC<Props> = ({ to }) => {
+  const preventDefault = (event: MouseEvent<HTMLButtonElement>) => event.preventDefault();
```
juanpicado: The user clicks the button and then what happens? .... We should open the target in a new tab.

juanpicado: We must also update `@verdaccio/types` to be able to have

```
  "funding": {
    "type": "opencollective",
    "url": "https://opencollective.com/verdaccio"
  },
```
in https://github.com/verdaccio/monorepo/blob/master/core/types/index.d.ts#L132 as optional type. Then can be used here.

juanpicado: You can find all the rules about that field here https://github.com/npm/rfcs/pull/54/files#diff-1bc199539d172d3211c84688a617e18cR30

juanpicado (commented):
Looks nice 👍

src/components/DetailSidebar/DetailSidebarFundButton.tsx
priscilawebdev: @juanpicado thank you for your feedback :) 

The component StyledLink has the prop "external". When this prop is set to true, the anchor element will have target="_blank" .😉

anikethsaha: so is it configurable from the verdaccio config file ?

I would make it default to new tab (\_blank)

juanpicado:
The best would be to have a PR with the whole thing instead spread the feature in multiple PR, this is helpful when we need to track issues or when features were introduced, otherwise the changelog file looks cleaner and anyone can identify the exact version where a feat was introduced.

DanielRuf:
Smaller commits / PRs make bisecting easier (I know, we squash them by default).

PR #399, opened 2019-12-25
fix: remove double padding and add missing background color - closes #373

### LINKED ISSUE #373: Background and container is duplicated
**Describe the bug**
There is a double container which add extra 15px padding to the whole container. Furthermore the background is not white `#FFFFF` anymore.

**To Reproduce**

Compare `yarn dev` `http://localhost:4872/-/web/detail/verdaccio` vs `https://registry.verdaccio.org/-/web/detail/verdaccio`

**Expected behavior**

Remove the extra-container

**Screenshots**
Before

Screen Shot 2019-12-16 at 9 13 50 PM

now

Screen Shot 2019-12-16 at 9 14 02 PM

DanielRuf:
**Type:**
bugfix

*  Prevent double padding by rmeoving it
*  Add missing background color

**Description:**

Resolves #373

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -17,7 +17,6 @@ import AppContextProvider from './AppContextProvider';
 import AppRoute, { history } from './AppRoute';
 
 const StyledBoxContent = styled(Box)<{ theme?: Theme }>(({ theme }) => ({
-  padding: 15,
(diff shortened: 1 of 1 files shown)


DanielRuf:
It seems there is an issue with some bot token and we have some test (and snapshot) which is time based (a year ago vs 2 years ago).

Issue #402, opened 2019-12-29
node >= 10.13.0 required


patsevanton:
Hello!

> Unsupported engine for @verdaccio/ui-theme@0.3.9: wanted node >= 10.13.0. Current node 8.17.0

cat /etc/redhat-release 
```CentOS Linux release 7.6.1810 (Core) ```

node -v
```v8.17.0```

npm install -g verdaccio
```
/usr/bin/verdaccio -> /usr/lib/node_modules/verdaccio/bin/verdaccio
npm WARN notsup Unsupported engine for @verdaccio/ui-theme@0.3.9: wanted: {"node":">= 10.13.0","npm":">=5"} (current: {"node":"8.17.0","npm":"6.13.4"})
npm WARN notsup Not compatible with your version of node/npm: @verdaccio/ui-theme@0.3.9

+ verdaccio@4.4.0
```

![image](https://user-images.githubusercontent.com/10828883/71556602-16c08280-2a65-11ea-9ad7-ce83bd4fc5c9.png)

DanielRuf:
I wonder why we don't test Node 8 anymore:

https://github.com/verdaccio/verdaccio/blob/master/.circleci/config.yml#L15

cc @juanpicado

DanielRuf:
https://github.com/verdaccio/verdaccio/pull/1495

cc @sergiohgz

DanielRuf:
So our docs are also outdated as we require npm >= 5 too.

https://verdaccio.org/docs/en/installation#prerequisites

patsevanton:
May be change https://verdaccio.org/docs/en/installation#prerequisites  ?

DanielRuf:
Mo, that change should not have happened actually as Node 10 requirement was planned for verdaccio 5.

We follow SemVer and verdaccio 4 itself requires Node >= 8. verdaccio 5 should require Node >= 10.

Also see https://github.com/verdaccio/verdaccio/pull/1335 and https://github.com/verdaccio/verdaccio/blob/master/package.json#L136

DanielRuf:
No, that change should not have happened actually as Node 10 requirement was planned for verdaccio 5.

We follow SemVer and verdaccio 4 itself requires Node >= 8. verdaccio 5 should require Node >= 10.

Also see https://github.com/verdaccio/verdaccio/pull/1335 and https://github.com/verdaccio/verdaccio/blob/master/package.json#L136

DanielRuf:
[...]

PR #403, opened 2019-12-30
fix: engine warning on console for ui

juanpicado:
**Type:** bug

*  There is a related issue?

**Description:**

```
"engines": {
    "node": ">= 10.13.0",
    "node": ">= 8",
    "npm": ">=5"
  },
```
We still support `v8` and this is a dependency of verdaccio.

Resolves https://github.com/verdaccio/ui/issues/402


DanielRuf:
@juanpicado doesn't `url-loader` 3 require Node >= 10? So this will still throw a warning or error.

juanpicado:
Yes, it does, but I think nobody is developing on Node 8` anyway. The change https://github.com/webpack-contrib/url-loader/pull/198 does not affect the library behavior.

This is just a dependency and we ship a webpack bundle.  I updated the README.

DanielRuf:
> This is just a dependency and we ship a webpack bundle. I updated the README.

You are right. I saw the other comment about the `devDependency` so this is not relevant for the prod bundles as you wrote =)

README.md
```diff
@@ -22,7 +22,7 @@
 
 ## Contributing
 
-We use `>=yarn@1.13.0`, keep on mind we use lock file.
+We use `>=yarn@1.13.0`, keep on mind we use lock file and use at least Node `v10.13.0` to be able to build the project.
```
DanielRuf:
```suggestion
We use `>=yarn@1.13.0`, keep in mind that we use lockfiles and use at least Node `v10.13.0` to be able to build the project.
```

PR #406, opened 2020-01-02
Fix: removed unused style file

### LINKED ISSUE #95: Investigate how to remove !important in styles
`!important` CSS keyword carries several behaviours in accessibility, so we have to try avoid their usage as much as we can.

We could try to make important things by repeating CSS class name instead, as commented in the PR referenced below.

_Originally posted by @sergiohgz in https://github.com/verdaccio/ui/pull/85_

priscilawebdev:
**Type:** Bug

**Description:**

Fixes #95


sergiohgz:
Did you add invalid for a reason?

priscilawebdev:
@sergiohgz yes... the code I removed doesn't seems right..hmm but actually it is a solution...I will remove the tag :)

sergiohgz:
If it doesn't seems right, does it really apply?

priscilawebdev:
sorry...I confused you..my PR solves the problem 😉

sergiohgz:
Aaaa, perfect 👌

PR #408, opened 2020-01-03
test: add SonarCloud

DanielRuf:
**Type:**
test / CI

*  prepare SonarCloud

**Description:**

This adds the required minimal configuration to enable SonarCloud scans.


juanpicado:
@DanielRuf app approved

DanielRuf:
Awesome. Now you just have to create the org and project at sonarcloud.io through https://github.com/apps/sonarcloud/installations/new for `verdaccio/ui`.

sergiohgz:
We need to configure to execute on a CI, in this case, I think CircleCI should be the target.

DanielRuf:
> We need to configure to execute on a CI, in this case, I think CircleCI should be the target.

We don't have to, it will be automatically added as extra check when it is completely setup and configured.

No need to change CI setup.

DanielRuf:
At least it should be only for PRs as extra check (code quality, code smells, ...).

juanpicado:
I will do some test

PR #412, opened 2020-01-08
feat: expand allowed protocols for URLs in repository component

### LINKED ISSUE #301: Suggestion: Perhaps support additional protocols in the package repository section
Right now only the following are protocols are classified as URL and displayed by the web interface:

https://github.com/verdaccio/ui/blob/b56e43846bc1492d15cd5c4e49857e4f1bfa4c1c/src/utils/url.ts#L7

Perhaps it would be useful for other protocols to be supported, specially all protocols that package.json supports. These being git, git+ssh, git+http, git+https and git+file.

Is there a particular reason for them to not be, specially because I think a fix for this would be very straightforward?

Edit: 
I've just checked the tests. https://github.com/verdaccio/ui/blob/5d300cd9be530197bc0eb4e413ad04bddc9cec68/src/components/Repository/Repository.test.tsx#L63

Is really this behavior wanted?

a666:
**feature:**

Allow other protocols in repository component

**Description:**

Resolves #301.

### DIFF
--- a/src/pages/Version/DetailSidebar/Repository/Repository.test.tsx
+++ b/src/pages/Version/DetailSidebar/Repository/Repository.test.tsx
@@ -57,7 +57,7 @@ describe('<Repository /> component', () => {
         ...detailContextValue.packageMeta?.latest,
         repository: {
           type: 'git',
-          url: 'git://github.com/verdaccio/ui.git',
+          url: 'invalid://github.com/verdaccio/ui.git',
         },
       },
     };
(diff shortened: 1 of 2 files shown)


DanielRuf (changes requested):

a666:
Sorry about the lack of updates, I just changed jobs and probably won't have the time to work on this feature. If someone else wants to have a go at it, they are welcome.

juanpicado:
Thanks for the update @a666 very appreciated, let's see whether someone wants to PR this PR :)

juanpicado (approved):
good enough for now, I'll add test on the main repo later on.

PR #415, opened 2020-01-09
fix: package list refresh based on logged-in user

### LINKED ISSUE #414: Failed to load all packages after login 2
**Describe the bug**
Non-public packages not included in the package-list after login.
A similar issue occurred in the past but the cause was different.

**To Reproduce**

1. Load the UI in a browser
2. You can see the public packages available to all
3. Login
4. Package list still the same, not included non-public packages I can access

**Expected behavior**
A package list containing all packages I can access.

> I have a PR ready solving this issue

coolsp:
**Type: fix**

**Description:**

In `pages/home/Home.tsx` now monitoring any change in a user log-in/out which will trigger a new `API.request` to get the _packages_ from the Verdaccio-server.  This is done by creating a `useEffect` on **isUserLoggedIn**.  Code has been transplanted from `App/App.tsx` including the use of the Loading component during the XHR.  The use of **packages** was removed from other components as no longer needed and tests updated.

Resolves #414


juanpicado (dismissed):
Thanks @coolsp good job

There is something we need to keep on mind and worth to document on the website. In our local dev config we have the following
```
security:
  api:
    jwt:
      sign:
        expiresIn: 120d
        notBefore: 1
  web:
    sign:
      expiresIn: 100d
      notBefore: 1
````
Let's focus on `notBefore: 1` for *web* section for a second. According the docs
https://github.com/auth0/node-jsonwebtoken#jwtsignpayload-secretorprivatekey-options-callback

> notBefore: expressed in seconds or a string describing a time span zeit/ms.
Eg: 60, "2 days", "10h", "7d". A numeric value is interpreted as a seconds count. If you use a string be sure you provide the time units (days, hours, etc), otherwise milliseconds unit is used by default ("120" is equal to "120ms").

If the user has something different than `notBefore: 0` it will get the following error in the console and the refresh won't work (which is the case of this repo https://github.com/verdaccio/ui/blob/master/tools/_verdaccio.config.yaml#L26).

Screen Shot 2020-01-12 at 6 23 59 PM

So, as I suggested, worth to document  it either 
* https://github.com/verdaccio/website/blob/master/docs/web.md
* https://github.com/verdaccio/website/blob/master/docs/config.md#security

juanpicado:
Please run `test:update-snapshot` and push again, snapshots are outdated.

coolsp (commented):
Test snapshots updated

juanpicado (approved):

juanpicado:
Thanks so much for fixing this!

Issue #401, opened 2019-12-28
Add script to update snapshots


DanielRuf:
**Is your feature request related to a problem? Please describe.**
As a developer I want to update snapshots when needed.

**Describe the solution you'd like**
As @juanpicado proposed:
> At jest you need to add an argument -u. Perhaps worth to add a new script
> 
> "test:snapshopts": "npm run test -- -u",
> 
> to facilitate this process while the other ticket is being fixed.

agar23:
is #416 all we need?

sergiohgz:
I don't think we need to create an scripts to update snapshots, as it's not a task that should bun run frequently.
If we still want a script, I think `test:snapshot` it's not semantically correct, it tells me more to run snapshot tests than update them.

agar23:
If we do decide to merge this, I updated the script to `test:update-snapshot`

PR #416, opened 2020-01-11
chore/401 add new script to update tests

agar23:
**Type:**

*  There is a related issue? #401 

**Description:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -171,7 +171,7 @@
     "test:acceptance:server": "concurrently --kill-others \"npm run verdaccio:server\" \"npm run test:acceptance\"",
     "test:e2e": "cross-env BABEL_ENV=test jest --config ./test/jest.config.e2e.js",
     "test": "cross-env NODE_ENV=test BABEL_ENV=test TZ=UTC jest --config ./jest/jest.config.js --maxWorkers 2 --passWithNoTests",
-    "test:snapshopts": "npm run test -- -u",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -171,6 +171,7 @@
     "test:acceptance:server": "concurrently --kill-others \"npm run verdaccio:server\" \"npm run test:acceptance\"",
     "test:e2e": "cross-env BABEL_ENV=test jest --config ./test/jest.config.e2e.js",
     "test": "cross-env NODE_ENV=test BABEL_ENV=test TZ=UTC jest --config ./jest/jest.config.js --maxWorkers 2 --passWithNoTests",
+    "test:snapshopts": "npm run test -- -u",
```
sergiohgz: The script name tells me more to run snapshot tests than update. Also, snapshoPts is misspelled.

agar23: make sense, updated to `test:update-snapshot`

juanpicado: I would rather go test:update and using yarn instead npm

DanielRuf: But we use `npm ...`in other scripts too. If this should be changed to `yarn`, shouldn't it be for all or for more?

sergiohgz: We should migrate all scripts to use one package manager, but that is other issue, let's open a new one and keep this simple :wink:

DanielRuf: Exacctly.

Issue #102, opened 2019-07-13
i18n at UI 


juanpicado:
**Is your feature request related to a problem? Please describe.**

Verdaccio is in English by default, but I want to change that. I would like to have multiple languages via configuration.

**Describe the solution you'd like**

I want UI (perhaps backend) in multiple languages. I would like to take advance of crowdin platform for this. 

**Describe alternatives you've considered**

I am not clear at this stage, perhaps

https://lingui.js.org/

Or similar things.

DanielRuf:
lingui seems a bit weird (we have to wrap the translatable strings in another component) https://lingui.js.org/ref/react.html#ref-react

https://github.com/i18next/react-i18next looks more familiar with translation implementations in other frameworks which use some `t()` or `_()` helper methods.

DanielRuf:
https://codesandbox.io/s/1zxox032q

juanpicado:
@DanielRuf seems nice, pretty much what I've seen in others by far. We need to run some test before choose anyone and see how might works with Crowdin.

juanpicado:
https://link.medium.com/08QfpgvAsZ

DanielRuf:
Intl is basically for helper methods as it formats dates, currency, time and a few more but is not intendend to implement i18n for a whole application.

Let's not reinvent the wheel and not do it from scratch.

juanpicado:
:-) sorry I did not have any better place to save the article 😊

DanielRuf:
Ah okay =)

juanpicado:
Intersting conversation and related https://github.com/gatsbyjs/rfcs/pull/42

juanpicado:
I will work on this.

juanpicado:
My strategy:

- fetch data from crowdin
- create some sort of data for UI
- prepare data for i18n react plugin 
- prepare build
- add new config param for chosen language
- build some test 
- cross fingers all works 
- new translations will be available to download only those has rights and periodically (later on we can automate this)
- I want crowdin be the central point for all translations

priscilawebdev:
Do you guys (@verdaccio/core-team) mind if I take over? I would love to implement i18n here

DanielRuf:
You can work on it @priscilawebdev =)

juanpicado:
@priscilawebdev ignore crowdin as I rewrote my thoughts above, find a different strategy, for now location files should be hosted on this repo.

PR #432, opened 2020-02-02
feat(i18n): added i18next for user interface translations 

priscilawebdev:
**Type:** Feat

The following has been addressed in the PR: #102 

*  Unit or Functional tests are included in the PR: 
Yes they are fixed, but it seems that I need to fix e2e tests now


i18n/config.ts
```diff
@@ -0,0 +1,29 @@
...
+import LanguageDetector from 'i18next-browser-languagedetector';
+import { initReactI18next } from 'react-i18next';
+
+i18n
+  // load translation using xhr -> see /i18n/translations
+  // learn more: https://github.com/i18next/i18next-xhr-backend
+  .use(Backend)
+  // detect user language
+  // learn more: https://github.com/i18next/i18next-browser-languageDetector
+  .use(LanguageDetector)
```
juanpicado: One reminder, the language must be changed vi a configuration rather via browser detection, you can check the description here https://github.com/verdaccio/ui/issues/102

juanpicado: So you need:
- [ ] Add new option https://github.com/verdaccio/verdaccio/blob/master/src/api/web/index.ts#L86
- [ ] Update documentation at website how to use this new param https://github.com/verdaccio/website/blob/master/docs/web.md
- [ ] The language must be taken via configuration and the UI render it

i18n/config.ts
```diff
@@ -0,0 +1,29 @@
+import i18n from 'i18next';
+import Backend from 'i18next-xhr-backend';
```
juanpicado: I'd not go would not go with this one, the backend is not ready to serve these files.

priscilawebdev: yeah I was going to remove it :)

i18n/config.ts
juanpicado: I am preparing a PR to simplifying the config passed to verdaccio to UI, I will include language so you can use real conf.

.eslintrc
```diff
@@ -96,7 +96,7 @@
     "react/jsx-pascal-case": ["error"],
     "react/jsx-props-no-multi-spaces": ["error"],
     "react/jsx-sort-default-props": ["error"],
-    "react/jsx-sort-props": ["error"],
+    "react/jsx-sort-props": 0,
```
juanpicado: 😉

i18n/config.ts
```diff
@@ -0,0 +1,31 @@
...
+import translationPT from './translations/pt-BR.json';
+
+i18n
+  // pass the i18n instance to react-i18next.
+  .use(initReactI18next)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    // in case window.VEDACCIO_LANGUAGE is undefined,it will fall back to 'en-US'
+    lng: window.VEDACCIO_LANGUAGE,
```
juanpicado: Here should be 
```
lng: window?.__VERDACCIO_BASENAME_UI_OPTIONS?.language,
```

juanpicado: and then update ...
Screen Shot 2020-03-03 at 7 47 38 AM

All new config comes from `__UI_OPTIONS`

src/App/AppRoute.tsx
```diff
@@ -36,36 +36,34 @@ const AppRoute: React.FC = () => {
 
   return (
     <Router history={history}>
-      <Suspense fallback={<Loading />}>
```
juanpicado: I need you to explain why you are moving `<Suspense/>`  from top-level to two different parts of the app, Header and App ?

src/components/Header/Header.tsx
```diff
@@ -41,7 +43,7 @@ const Header: React.FC<Props> = ({ withoutSearch }) => {
   };
 
   return (
-    <>
+    <Suspense fallback="loading">
```
juanpicado: I guess you don't want `fallback="loading"` to display a string.

src/components/Label/Label.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable verdaccio/jsx-spread*/
```
juanpicado: 🤔

src/components/NotFound/Notfound.test.tsx
```diff
@@ -14,17 +14,16 @@ describe('<NotFound /> component', () => {
     );
     expect(container.firstChild).toMatchSnapshot();
   });
-  test('go to Home Page button click', () => {
-    const spy = jest.spyOn(React, 'useCallback');
-    const { getByText } = render(
+
+  // TODO (priscila): fix this test
+  test.skip('go to Home Page button click', async () => {
```
juanpicado: TODO 👀

src/components/Package/Package.tsx
```diff
@@ -155,7 +159,7 @@ const Package: React.FC<PackageInterface> = ({
             <PackageTitle className="package-title">{packageName}</PackageTitle>
           </WrapperLink>
         </Grid>
-        <GridRightAligned item={true} xs={true}>
+        <GridRightAligned alignItems="center" container={true} item={true} justify="flex-end" xs={true}>
```
juanpicado: Does this PR requires change this `alignItems="center" container={true} item={true} justify="flex-end"``?

src/icons/GitHub.tsx
```diff
@@ -1,14 +0,0 @@
-/* eslint-disable max-len */
```
juanpicado: Why this is being removed?

src/utils/login.test.ts
```diff
@@ -16,6 +16,22 @@ jest.mock('./api', () => ({
   request: require('../../jest/unit/components/__mocks__/api').default.request,
 }));
 
+jest.mock('i18next', () => {
+  const translationEN = require('../../i18n/translations/en-US.json');
```
juanpicado: Do we need this? I don't see the connection with the test.

tools/dev.server.js
```diff
@@ -17,7 +17,7 @@ compiler.hooks.done.tap('Verdaccio Dev Server', () => {
 });
 
 new WebpackDevServer(compiler, {
-  contentBase: `${env.DIST_PATH}`,
```
juanpicado: Why this?

types/index.ts
```diff
@@ -12,5 +12,6 @@ declare global {
     VERDACCIO_SCOPE: string;
     VERDACCIO_VERSION: string;
     VERDACCIO_API_URL: string;
+    VEDACCIO_LANGUAGE: string;
```
juanpicado: Reminder, this should be part of `__VERDACCIO_BASENAME_UI_OPTIONS` and `VerdaccioOptions` as `language`.

src/App/AppRoute.tsx
juanpicado: Ok, I think you forgot to remove the `<Suspense/>` from `Header`.

i18n/config.ts
priscilawebdev: oh yeah...I forgot to add back the English language

src/icons/GitHub.tsx
priscilawebdev: I may be mistaken, but I couldn't find where it is being used

src/components/Label/Label.tsx
priscilawebdev: I was getting an warning and decided to add this `eslint-disable,` but I have just removed it :)

image

src/components/Package/Package.tsx
priscilawebdev: yes! it was necessary to do this changes

src/utils/login.test.ts
priscilawebdev: in fact they have a connection ... I'm using the English translation file to do the tests ... I didn't want to write consts or a string ... I think it's better if we reuse it ... so let's say I don't I want more "Login", but "Sign in" .. if the id's remained the same (en translations) and only the value changes, we don't need to update the tests

tools/dev.server.js
priscilawebdev: we don't need to wrap it in a string literal

priscilawebdev:
@juanpicado yes, some  words are not being translated, because either they are come from backend as a value, or from a library and I need to fix it ...please see:

image

juanpicado:
> @juanpicado yes, some words are not being translated, because either they are come from backend as a value, like "JS foundation and contributors", or from a library, '2 years ago' and I need to fix it ...please see:
> 
> image

ok, if comes from the backend, we will fix that later, thanks for letting me know :)

PR #438, opened 2020-03-04
Revert "chore: add snyk as an action"

anikethsaha:
Reverts verdaccio/ui#230

Github action doesn't passes the tokens to forks so this cant be tested against them.

It is recommended to use the snyk webhooks.


juanpicado:
We had a chat about this @anikethsaha , so feel free to close it.

Issue #427, opened 2020-01-28
Enable the theme to be customized 🌟💅


priscilawebdev:
**Is your feature request related to a problem?**
Not really

**Describe the solution you'd like**
I believe that the possibility of customizing the logo and colors is one of the "most wanted" features of the UI. For the happiness of our users, we decided to include this feature in the next release 🎉🚀

The idea is to have a type of configuration file, with the same properties that we have in the `theme.ts,` accepted for the logo. So it will be possible to set things like `primaryColor, secondaryColor, fontSize`, etc

 Some research is still needed and suggestions are very welcome!👏

juanpicado:
I did a quick try, just a PoC https://github.com/verdaccio/ui/tree/feat-dark-mode and I found out change the colors is not that easy, but doable.

priscilawebdev:
@juanpicado really nice PoC :)

I just find the background too dark hahaha

![image](https://user-images.githubusercontent.com/29228205/76709830-2f564580-6702-11ea-83ac-3c0959091b9b.png)

juanpicado:
@priscilawebdev should we have this open ? seems like you made it already.

priscilawebdev:
@juanpicado , I created a dark theme, but I still wanted to make the theme more customizable ... let's close it for now :)

PR #446, opened 2020-03-22
feat(style): added dark mode

priscilawebdev:
**Type:** Feature

**Description:** Added Dark Mode Theme

dark-mode

Step 01 of https://github.com/verdaccio/ui/issues/427


src/design-tokens/ThemeProvider.tsx
```diff
@@ -1,13 +1,26 @@
...
+import ThemeContext from './ThemeContext';
+import { getTheme, ThemeMode } from './theme';
 
-const ThemeProvider: React.FC = ({ children }) => (
-  <EmotionThemeProvider theme={theme}>
-    <MuiThemeProvider theme={theme}>{children}</MuiThemeProvider>
-  </EmotionThemeProvider>
-);
+const ThemeProvider: React.FC = ({ children }) => {
+  const [isDarkMode, setIsDarkMode] = useState();
```
juanpicado: If users want to have dark mode by default, how we can achieve that right now? I think this might be the entry point using `UI_OPTIONS` as we did with language, by default would be the default theme.

priscilawebdev: done 😉

types/index.ts
```diff
@@ -3,7 +3,7 @@ export interface VerdaccioOptions {
   url_prefix: string;
   base: string;
   language?: string;
-  darkMode?: boolean;
+  dark_mode?: boolean;
```
juanpicado: I prefer cammelCase, hopefully we can remove some day all the `_` ...

priscilawebdev: me too, but then I noticed that we have `url_prefix` and I decided to do like this

juanpicado: `url_prefix` is from the *sinopia* era, the previous author was quite all over the place using that, it took me few years to get rid of those, but some configurations are still using `_` due backward compatibility unfortunately.

src/design-tokens/ThemeProvider.tsx
```diff
@@ -1,13 +1,27 @@
...
+import { getTheme, ThemeMode } from './theme';
 
-const ThemeProvider: React.FC = ({ children }) => (
-  <EmotionThemeProvider theme={theme}>
-    <MuiThemeProvider theme={theme}>{children}</MuiThemeProvider>
-  </EmotionThemeProvider>
-);
+const ThemeProvider: React.FC = ({ children }) => {
+  const isDarkModeDefault = window?.__VERDACCIO_BASENAME_UI_OPTIONS?.dark_mode;
+  const [isDarkMode, setIsDarkMode] = useState(!!isDarkModeDefault);
```
juanpicado: This works really well, but we might use the local storage to persist in the dark mode. Let's imagine the following scenario:
* Admin set `dark_mode` as `true` 
* Dark mode is set by default for all users.
* But here is one which likes green or primary color.
* This user will try to persist primary theme but it is not possible right now, on refresh it will switch again.

priscilawebdev: Yes I noticed this problem and I was thinking to add session storage or local storage, but maybe I could do i in another PR?

juanpicado:
@priscilawebdev updated https://github.com/verdaccio/website/blob/master/docs/web.md#configuration feel free to merge :)

PR #456, opened 2020-04-05
feat(lng): Added change language on the fly

priscilawebdev:
Change Language on the fly

image

image


src/components/Icon/Icon.tsx
```diff
@@ -11,6 +11,7 @@ import nicaragua from './img/nicaragua.svg';
 import pakistan from './img/pakistan.svg';
 import austria from './img/austria.svg';
 import spain from './img/spain.svg';
+import usa from './img/usa.svg';
```
juanpicado, line 14: USA? 🙃

src/components/LanguageSwitch/LanguageSwitch.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const { t } = useTranslation();
+  const [open, setOpen] = React.useState(false);
+  const anchorRef = React.useRef<HTMLButtonElement>(null);
+
+  const languages = i18next.options.resources ? Object.keys(i18next.options.resources) : [];
+  const currentLanguage = i18next.language || i18next.options?.fallbackLng?.[0];
+
+  const { translation: userLanguage } = getTranslatedCurrentLanguage(t)[currentLanguage.toLowerCase()];
+
+  const handleToggle = () => {
```
juanpicado, line 58: we might optimize the callbacks with `useCallback`, what do you think?

src/components/LanguageSwitch/LanguageSwitch.tsx
```diff
@@ -0,0 +1,157 @@
...
+                        <Icon name={icon} size="md" />
+                        {translation}
+                      </StyledMenuItem>
+                    );
+                  })}
+                  <Box my={1}>
+                    <Divider />
+                  </Box>
+                  <MenuItem button={true}>
+                    <StyledLink external={true} onClick={handleClose} to="https://github.com/verdaccio/ui#translations">
```
juanpicado, line 120: I'd add this to a constant above, it is hardly visible here `https://github.com/verdaccio/ui#translations` . Furthermore, the README might change and this link will be broken. I'd rather remove the `#translations` part to be sure.

src/components/Icon/Icon.tsx
priscilawebdev: yes :) United States of America

src/components/LanguageSwitch/LanguageSwitch.tsx
priscilawebdev: I think it's valid :)

juanpicado:
I just merged French, I think we need also to update it here.

priscilawebdev:
@juanpicado thanks for the nice feedback! I just added the current language to localStorage, so it should look good now if you reload the page ...

I also updated the menu background in dark mode 😉

Regarding the header on mobile devices, I plan to refactor it by adding a mobile menu, because atm it is impossible to add another icon / information ... as you said, it looks odd ... I will do this in a follow-up PR ... but in the meantime, the component switchLanguage will be hidden on smaller devices 

I'm not having much time lately, but I will try to do that as soon as I have some free time :)

juanpicado:
> @juanpicado thanks for the nice feedback! I just added the current language to localStorage, so it should be fine now if you reload the page ...
> 
> I also updated the menu background color in dark mode 😉
> 
> Regarding the header on mobile devices, I plan to refactor it by adding a mobile menu, because atm it is impossible to add another icon / information ... as you said, it looks odd ... I will do this in a follow-up PR ... but in the meantime, the component switchLanguage will be hidden on smaller devices
> 
> I'm not having much time lately, but I will try to do that as soon as I have some free time 🙂

Perfect, good solution 👍

PR #457, opened 2020-04-07
fix: when scope is provided use proper separator

JGjorgji:
**Type:** Fix

**Description:**

it's not documented particularly well but ":" needs to be set between
the scope and "registry" in order to set the proper .npmrc config

https://docs.npmjs.com/files/npmrc#comments

this demonstrates it well although the link is for a different topic

this was incorrectly fixed in https://github.com/verdaccio/ui/pull/34


JGjorgji:
> Thanks for this PR, would be great to have unit test for this pray

Added some unit tests, let me know if they're ok.

src/utils/cli-utils.ts
```diff
@@ -21,7 +21,8 @@ export const copyToClipBoardUtility = (str: string): ((e: SyntheticEvent<HTMLEle
 };
 
 export function getCLISetConfigRegistry(command: string, scope: string, registryUrl: string): string {
-  return `${command} ${scope}registry ${registryUrl}`;
+  // if there is a scope defined there needs to be a ":" separator between the scope and the registry
+  return `${command} ${scope}${scope !== '' ? ':' : ''}registry ${registryUrl}`;
```
priscilawebdev, line 25:
```suggestion
  return `${command} ${scope}${!!scope ? ':' : ''}registry ${registryUrl}`;
```

priscilawebdev: I believe registry should be translated

priscilawebdev:
@JGjorgji thanks for this great PR. I left just some comments about some minor changes 😉

src/utils/cli-utils.ts
juanpicado: I think registry cannot be translated, it is a command. It has to be in English.

priscilawebdev: oh sure! I missed it ..thanks @juanpicado 👍

src/utils/cli-utils.ts
```diff
@@ -21,7 +21,8 @@ export const copyToClipBoardUtility = (str: string): ((e: SyntheticEvent<HTMLEle
 };
 
 export function getCLISetConfigRegistry(command: string, scope: string, registryUrl: string): string {
-  return `${command} ${scope}registry ${registryUrl}`;
+  // if there is a scope defined there needs to be a ":" separator between the scope and the registry
+  return `${command} ${scope}${!!scope ? ':' : ''}registry ${registryUrl}`;
```
priscilawebdev, line 25:
```suggestion
 return `${command} ${scope}${scope !== '' ? ':' : ''}registry ${registryUrl}`;
```

Issue #461, opened 2020-04-24
Current locale breaks UI


juanpicado:
**Describe the bug**
More detail here https://github.com/verdaccio/verdaccio/issues/1779

We assume the `currentLanguage` will be the locale of the browser, but not everybody has the same locale, this is mine.

<img width="366" alt="Screen Shot 2020-04-24 at 6 12 01 PM" src="https://user-images.githubusercontent.com/558752/80233676-2e94c580-8657-11ea-9a8b-eedd6ba78d4e.png">

And that's not on the list.

Diving into `i18next`

<img width="335" alt="Screen Shot 2020-04-24 at 6 14 18 PM" src="https://user-images.githubusercontent.com/558752/80233862-77e51500-8657-11ea-80a1-55f909bfa608.png">

Solution is here https://www.i18next.com/overview/api#language

```
const VERDACCIO_UI_GITHUB_REPOSITORY = 'https://github.com/verdaccio/ui';

const getTranslatedCurrentLanguage = (
  t: TFunction
): { [key: string]: { translation: string; icon: React.ComponentProps<typeof Icon>['name'] } } => ({
  'en-us': {
    translation: t('lng.english'),
    icon: 'usa',
  },
  'fr-fr': {
    translation: t('lng.french'),
    icon: 'france',
  },
  'pt-br': {
    translation: t('lng.portuguese'),
    icon: 'brazil',
  },
  'de-de': {
    translation: t('lng.german'),
    icon: 'germany',
  },
  'es-es': {
    translation: t('lng.spanish'),
    icon: 'spain',
  },
  'zh-cn': {
    translation: t('lng.chinese'),
    icon: 'china',
  },
});
```

We define all languages as lower case.

**Screenshots**
<img width="639" alt="Screen Shot 2020-04-24 at 6 04 34 PM" src="https://user-images.githubusercontent.com/558752/80232996-13758600-8656-11ea-94fc-0ce7d5ce9bb1.png">
<img width="611" alt="Screen Shot 2020-04-24 at 6 03 00 PM" src="https://user-images.githubusercontent.com/558752/80233005-17a1a380-8656-11ea-9155-7b1ed31bcd8a.png">

juanpicado:
On development does not fail, but it does in production. The object `i18next.languages` does not exist locally/dev.

PR #462, opened 2020-04-26
fix(i18n): fixed current locale

priscilawebdev:
**Type:** Fix

The following has been addressed in the PR: https://github.com/verdaccio/ui/issues/461

*  There is a related issue? Yes

**Description:**

Closes https://github.com/verdaccio/ui/issues/461 & added missing Japanese translations


i18n/translations/ja-JP.json
```diff
@@ -135,5 +135,16 @@
...
-}
\ No newline at end of file
+  },
+  "lng": {
+    "english": "英語",
+    "japanese": "日本語",
+    "portuguese": "ポルトガル語",
+    "spanish": "スペイン語",
+    "german": "ドイツ人",
+    "chinese": "中国の",
```
hdmr14, line 145:
```suggestion
    "german": "ドイツ語",
    "chinese": "中国語",
```

These lines should fixed like above.

priscilawebdev:
@hdmr14 ありがとうございました 🙏

PR #466, opened 2020-05-01
Update zh-CN.json

jfthuong:
Improve list of languages in Chinese (Simplified) and other misc improvements

**Type:**

*  There is a related issue?

**Description:**

### DIFF
--- a/i18n/translations/zh-CN.json
+++ b/i18n/translations/zh-CN.json
@@ -109,7 +109,7 @@
             "title": "贡献者"
         },
         "engines": {
-            "npm-version": "NPM Version",
+            "npm-version": "NPM版本",
             "node-js": "NODE JS"
         }
     },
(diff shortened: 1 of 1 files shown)


jfthuong:
There were few differences

1. 语 (language) was in traditional characters (which is used for example in Hong Kong) while the rest is using simplified characters (like it is used in mainland China)
2. "French" was translated as "French (person)" 😂
3. Sometimes, for language it was using "语" and sometimes "文" (which are synonyms). I have made it uniform

By the way, great devops tools (sonarcloud and codecov)!

jfthuong:
There were few differences

1. 语 (language) was in traditional characters (which is used for example in Hong Kong) while the rest is using simplified characters (like it is used in mainland China)
2. "French" was translated as "French (person)" 😂
3. Sometimes, for language it was using "语" and sometimes "文" (which are almost synonyms). I have made it uniform. Actually, a Chinese friend just helped me check and "语" is more for oral language while "文" is for written language... So all the 语 should be replaced by 文😅

By the way, great devops tools (sonarcloud and codecov)!

juanpicado:
Thanks again @jfthuong 🚀

PR #471, opened 2020-05-02
fix: time format localization

hdmr14:
**Type: translate**

* fix, the time format is not localized at package list.

*  There is a related issue?
No
Yes, It bit fixed due to time format changed

**Description:**
Verdaccio configure `TIME FORMAT` to `DD.MM.YYYY, HH:mm:ss`, but actually the date time format is strongly depend to language.

ex(day.js localization) ...
in English: `MM/DD/YYYY HH:MM:SS A`
in French: `DD/MM/YYYY HH:MM:SS`
in German: `DD.MM.YYYY HH:MM:SS`
in Japanese: `YYYY/MM/DD HH:MM:SS`

day.js has [localized format plugin](https://day.js.org/docs/en/plugin/localized-format) which solve date time format localization problem like above.

In this PR, use it to localize date time in package list and also contains fix tests for these changes.

### DIFF
--- a/src/utils/package.test.ts
+++ b/src/utils/package.test.ts
@@ -50,7 +50,7 @@ describe('formatRepository', (): void => {
 describe('formatDate', (): void => {
   test('should format the date', (): void => {
     const date = 1532211072138;
-    expect(formatDate(date)).toEqual('21.07.2018, 22:11:12');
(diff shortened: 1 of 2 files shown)


priscilawebdev:
@hdmr14 great job! thank you!!! 🙏 😊

PR #477, opened 2020-05-10
ref(ui): Updated autocomplete

priscilawebdev:
Preview:

Kapture 2020-05-23 at 16 26 54

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -85,7 +85,7 @@
     "react/jsx-key": ["error"],
     "react/jsx-max-depth":["error", { "max": 5}],
     "react/jsx-max-props-per-line": ["error", {"maximum": 3, "when": "multiline" }],
-    "react/jsx-no-bind": ["error"],
+    "react/jsx-no-bind": ["error", {"allowArrowFunctions": true}],
     "react/jsx-no-comment-textnodes": ["error"],
     "react/jsx-no-duplicate-props": ["error"],
     "react/jsx-no-literals": ["error"],
--- a/src/components/AutoComplete/AutoCompleteV2.tsx
+++ b/src/components/AutoComplete/AutoCompleteV2.tsx
@@ -1,0 +1,276 @@
+import React, { memo, useState, ChangeEvent, KeyboardEvent, useEffect, useRef, useCallback } from 'react';
+import styled from '@emotion/styled';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import CloseIcon from '@material-ui/icons/Close';
+import { withStyles } from '@material-ui/core/styles';
+import { useTranslation } from 'react-i18next';
+
+import TextField from '../../muiComponents/TextField';
+import Paper from '../../muiComponents/Paper';
+import MenuItem from '../../muiComponents/MenuItem';
+import IconButton from '../../muiComponents/IconButton';
+import { Theme } from '../../design-tokens/theme';
+import { useOnClickOutside } from '../../design-tokens/useOnClickOutside';
+
+import { createFilterOptions } from './useAutoComplete';
+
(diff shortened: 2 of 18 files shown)


.eslintrc
```diff
@@ -85,7 +85,6 @@
     "react/jsx-key": ["error"],
     "react/jsx-max-depth":["error", { "max": 5}],
     "react/jsx-max-props-per-line": ["error", {"maximum": 3, "when": "multiline" }],
-    "react/jsx-no-bind": ["error"],
```
priscilawebdev, line 88: Sometimes, it doesn't make sense to create an extra function. example: 

`() => setShowSuggestions(true)`

I would prefer a simple version instead of

```
handleSetShowSuggestons = () => {
 setShowSuggestions(true)
}
```

src/components/LanguageSwitch/LanguageSwitch.tsx
```diff
@@ -1,28 +1,16 @@
...
 import MenuItem from '../../muiComponents/MenuItem';
-import Button from '../../muiComponents/Button';
-import Tooltip from '../../muiComponents/Tooltip';
-import Divider from '../../muiComponents/Divider';
-import Box from '../../muiComponents/Box';
 import { Theme } from '../../design-tokens/theme';
-import Link from '../Link';
 import Icon from '../Icon';
-
-const VERDACCIO_UI_GITHUB_REPOSITORY = 'https://github.com/verdaccio/ui';
```
priscilawebdev, line 25: Now that we have  the "translation" section in our docs, I believe it is fine to remove it

.eslintrc
juanpicado: Could you give me a real example? Usually `useCallback` worth, others do not.

src/components/LanguageSwitch/LanguageSwitch.tsx
juanpicado: Ok to me.

juanpicado:
I think snapshopts are failing.

.eslintrc
priscilawebdev: https://github.com/verdaccio/ui/blob/5592f107a1296a90ae42b76606c8d24955be804f/src/components/LanguageSwitch/LanguageSwitch.tsx#L92

priscilawebdev:
Kapture 2020-05-23 at 16 26 54

.eslintrc
juanpicado: I see, well, in that case is ok. But just few lines below you might improve the performance of your component memoazing this

```
        options={languages.map(language => {
          const { icon, translation } = getTranslatedCurrentLanguageDetails(t, language);
          return {
            language,
            icon,
            translation,
          };
        })}
```
Languages hardly will change  so this is an scenario where makes totally sense using `useMemo` .

juanpicado: I'd restore the rule, if sometimes does not make sense, you always can disable the rule for such line. In the worse case, you can customize it https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#rule-options and make it more flexible.

src/components/AutoComplete/AutoCompleteV2.tsx
```diff
@@ -0,0 +1,276 @@
+import React, { memo, useState, ChangeEvent, KeyboardEvent, useEffect, useRef, useCallback } from 'react';
```
juanpicado, line 1: I'd rename this to something more close to the job was that, like ... `AutoCompleteDropwDown` or something like that.

priscilawebdev: sure thing!  I will do this in a follow-up PR ;-) 

but my idea here was to replace the current Autocomplete that we have..

PR #435, opened 2020-02-28
chore: html linter

juanpicado:
**Type:** build

**Description:**

@anikethsaha is proposing this linter made by him. I like the idea. Let's give it a try.

PR #479, opened 2020-05-24
Added htmllinter for linting html file

anikethsaha:
**Type:** build

- There is a related issue? not issue, PR https://github.com/verdaccio/ui/pull/435
- Unit or Functional tests are included in the PR No

**Description:**
I dont have the permission to push to this repo's `chore-html-linter` branch so I checkout that branch and made changes on top of that to keep @juanpicado as co-author.

Currently This linting for HTML will fail because there is an empty tag in the template so that will violate [`no-empty-tag`](https://github.com/anikethsaha/htmllinter/blob/master/packages/basic-rules/src/rules/no-empty-tag.js) rule.
`htmllinter` doesn't support auto fix as of now.

So one fix would be to add `div` to rule's options like this

```js
// htmllinter.config.js
{
   rule: {
     "no-empty-tag": ['on', { ignore: ['div']}]
 }
}
```

Or another fix would be to wait for https://github.com/anikethsaha/htmllinter/issues/26 to land and release. 

IMO, I would wait for the second one as I will do it pretty soon. 😄 

***

There is also a [webpack plugin for this linter](https://github.com/anikethsaha/htmllinter/tree/master/packages/htmllinter-webpack-plugin) which will lint the HTML template before emitting and wont exit with any non-zero exit code. 

Let me know if you guys want that instead.


priscilawebdev:
@anikethsaha your package looks great 👏🌟 Please let us know when the second option will be available, so then we can merge this PR 😉

anikethsaha:
Thanks 😃 

I will fix that in this week for sure 👍

package.json
```diff
@@ -19,6 +19,8 @@
     "@commitlint/config-conventional": "8.3.4",
     "@emotion/core": "10.0.22",
     "@emotion/styled": "10.0.23",
+    "@htmllinter/basic-config": "^0.5.0",
```
juanpicado, line 22: let's pin here :-)

anikethsaha:
not sure why the canary CI is failing.

juanpicado:
> not sure why the canary CI is failing.

Not your fault, problem is GitHub secrets does not work with forks ...

anikethsaha:
> > not sure why the canary CI is failing.
> 
> Not your fault, problem is GitHub secrets does not work with forks ...

ohh , cool.

PR #497, opened 2020-06-14
ref(structure): Update folder structure 

priscilawebdev:
**Type:** Refactor

**Description:**

. Moves common components to the `src/components` folder
. Moves the components used in a specific location (1x), to the specific location.
. Create the alias `@verdaccio/components/* @verdaccio/design-tokens/* @verdaccio/utils/*`
. Update the dependency "eslint-plugin-import" in order to use https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#pathgroupsexcludedimporttypes-array
 
  - All the @verdaccio/* imports will be placed after the external imports.


jest/jest.config.js
```diff
@@ -26,5 +26,8 @@ module.exports = {
     'github-markdown-css': '<rootDir>/node_modules/identity-obj-proxy',
     '\\.(png)$': '<rootDir>/node_modules/identity-obj-proxy',
     '\\.(svg)$': '<rootDir>/jest/unit/empty.ts',
+    '@verdaccio/components/(.*)': '<rootDir>/src/components/$1',
```
priscilawebdev, line 29: I'd like to create a common file with these aliases, but I don't know exactly where to add it...Please, any thoughts?

PR #500, opened 2020-06-21
chore: add issue template config

priscilawebdev:
**Type:** Feat

**Description:**

Added issue template config


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,14 @@
...
+    url: https://stackoverflow.com/questions/tagged/verdaccio
+    about: We encorage you to open new questions on Stackoverflow
+  - name: Security Report 🛡
+    url: https://github.com/verdaccio/ui/security/policy
+    about: I want to report a security vulnerability
+  - name: Chat 🏘
+    url: http://chat.verdaccio.org
+    about: For a quick question you should do it through our community chat
+  - name: User Interface Report 👩🏼‍🎨👨🏼‍🎨
+    url: https://github.com/verdaccio/ui/issues/new/choose
```
juanpicado, line 13: Here you are pointing your self ;-) .. infinite loop. 

I'd rather modify the one we had in Verdaccio more adapted to the UI requirements. Furthermore this file itself brings not much value if we don't customize the other templates.

For instance, you might point out users to report bugs **non related** with the UI to Verdaccio repo. The bugs report should contains the requirements for UI which are relevant, like browser or specific steps to reproduce.

This is the entry point of the user, a well structure entry point will reduce the noise on non related questions with this repo 🙃 and improve experience for everybody.

juanpicado: One little thing, this is not accurate https://github.com/verdaccio/ui/security/policy 
We have no version 5 yet and also duplicates the Verdaccio core policy, people might get confused, the better is either, redirect to verdaccio core (otherwise will be out of sync eventually) or just don't use it here at all.  We might ask @DanielRuf or @lirantal since they have more experience on this area.

PR #502, opened 2020-06-23
chore: increased explicit typings

tmkn:
**Description:**
This PR further increases compatibility with TypeScripts `strict` setting.
No logic changes in the code where necessary.

After this PR only `api.ts` and `AutoComplete.tsx` remain to be fixed but those require most likely code changes so I've left them out of this PR.
Aka after this PR only 14 errors remain to be fixed to enable `strict` mode. I remember the last time I worked on it that number was much higher 👍


src/utils/cli-utils.test.ts
```diff
@@ -3,7 +3,7 @@ import { SyntheticEvent } from 'react';
 import { copyToClipBoardUtility, getCLISetConfigRegistry } from './cli-utils';
 
 describe('copyToClipBoardUtility', () => {
-  let originalGetSelection;
+  let originalGetSelection: typeof window.getSelection;
```
tmkn, line 6: Maybe also a good idea to use `jest.spyOn` to test the clipboard in the future so it's not needed to reset `window.getSelection` manually, like it's done now. But as said the focus for this PR was to do no actual code changes, so I've left it untouched.

juanpicado: agree

PR #503, opened 2020-06-25
Feature: Strict Typing for AutoComplete Component

tmkn:
This PR makes `AutoComplete.tsx` compatible with TypeScripts `strict` mode.
Unfortunately I had to introduce a `@ts-ignore` directive in 1 place since the provided `RenderInputComponent<T>` type does not allow to specify custom input props which we make use of. See here: https://github.com/moroshko/react-autosuggest#input-props-prop

The first generic in `RenderInputComponent<T>` only models the Suggestion, so there is currently no way to type custom input props. Ideally in the future it will allow a 2nd generic that allows you to model them like `RenderInputComponent<Suggestion, CustomInputProps>`.
I already specified the custom input prop that would need to be used: `type CustomInputProps = Pick<Props, 'disableUnderline' | 'startAdornment'>;`
So once this is fixed we can remove the tsignore again and fill in the 2nd generic and this should be it.

### DIFF
--- a/src/components/AutoComplete/AutoComplete.tsx
+++ b/src/components/AutoComplete/AutoComplete.tsx
@@ -2,7 +2,7 @@ import styled from '@emotion/styled';
 import match from 'autosuggest-highlight/match';
 import parse from 'autosuggest-highlight/parse';
 import React, { KeyboardEvent, memo } from 'react';
-import Autosuggest, { SuggestionSelectedEventData, InputProps, ChangeEvent } from 'react-autosuggest';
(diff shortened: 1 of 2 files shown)


src/components/AutoComplete/AutoComplete.tsx
```diff
@@ -30,22 +39,29 @@ interface Props {
...
+  name: string;
+}
+
+type CustomInputProps = Pick<Props, 'disableUnderline' | 'startAdornment'>;
+
 /* eslint-disable react/jsx-sort-props  */
 /* eslint-disable verdaccio/jsx-spread */
-const renderInputComponent = (inputProps): JSX.Element => {
+const renderInputComponent: RenderInputComponent<Suggestion> = inputProps => {
+  // @ts-ignore
```
tmkn, line 58: this is the new ts-ignore, since there is currently no way to type `startAdornment` & `disableUnderline`

juanpicado: Ok, we can find a solution later for this

PR #504, opened 2020-06-29
Refactor: Enabled typescript strict mode

tmkn:
This PR removes the `"noImplicitAny": false` option from `tsconfig.json`.
Meaning the source code is now compatible with the Typescript `strict` setting, providing the highest type safety. 🎉 

To enable this I had to slightly adapt the ajax code, since it wasn't possible to type the existing code without introducing too many changes.
`API.request` can return a string, json or a blob based on the response content type or url ending or void if those checks don't match.
This proved to be a quite difficult scenario to type, so as to avoid too many changes to the existing code, `API.request` now returns a string (the response bodys string) where it would have previously returned `undefined`. This allowed for only minimal changes to the existing ajax code.

However the type safety in this part is not as good as it could be. e.g:
```typescript
const response: LoginBody = await API.request('login', ...);
```
While the code expects a `LoginBody`, at runtime it could return a Blob since as said the actual return type depends on the responses content type or url ending.
It might make sense to provide separate ajax functions for when to expect string, json or Blob like:
```typescript
const response: LoginBody = await API.requestJson('login', ...);
```
This way `API.requestJson` could throw when it can't return a valid json/finds a Blob. This would make the code more failsafe.
Anyway the current code nevertheless is now `strict` mode compatible.

I also removed all the temporary helper files that where added to incrementally make `strict` mode happen.

TypeScript was also updated to the (as of now) latest version: 3.9.5

tldr:

- Set `strict` mode compilation as the new default
- `API.request` now returns a `string` in cases where it would have returned `undefined` previously
- Updated Typescript from 3.7.4 to 3.9.5

### DIFF
--- a/src/App/Header/Header.tsx
+++ b/src/App/Header/Header.tsx
@@ -22,9 +22,9 @@ interface Props {
 const Header: React.FC<Props> = ({ withoutSearch }) => {
   const { t } = useTranslation();
   const appContext = useContext(AppContext);
-  const [isInfoDialogOpen, setOpenInfoDialog] = useState();
-  const [showMobileNavBar, setShowMobileNavBar] = useState();
-  const [showLoginModal, setShowLoginModal] = useState(false);
+  const [isInfoDialogOpen, setOpenInfoDialog] = useState<boolean>(false);
+  const [showMobileNavBar, setShowMobileNavBar] = useState<boolean>(false);
+  const [showLoginModal, setShowLoginModal] = useState<boolean>(false);
 
   if (!appContext) {
     throw Error(t('app-context-not-correct-used'));
--- a/src/utils/api.test.ts
+++ b/src/utils/api.test.ts
@@ -3,16 +3,18 @@ import api, { handleResponseType } from '../../src/utils/api';
 describe('api', () => {
   describe('handleResponseType', () => {
     test('should handle missing Content-Type', async () => {
+      const responseText = `responseText`;
+      const ok = false;
       const response: Response = {
         url: 'http://localhost:8080/-/packages',
-        ok: false,
+        ok: ok,
         headers: new Headers(),
+        text: async () => responseText,
       } as Response;
 
       const handled = await handleResponseType(response);
 
-      // Should this actually return [false, null] ?
-      expect(handled).toBeUndefined();
+      expect(handled).toEqual([ok, responseText]);
     });
 
     test('should test tgz scenario', async () => {
@@ -94,9 +96,9 @@ describe('api', () => {
 
       expect(fetchSpy).toHaveBeenCalledWith('https://verdaccio.tld/resource', {
         credentials: 'same-origin',
-        headers: {
+        headers: new Headers({
           Authorization: 'Bearer token-xx-xx-xx',
-        },
+        }),
         method: 'GET',
       });
       expect(response).toEqual({ c: 3 });
--- a/src/utils/api.test.ts
+++ b/src/utils/api.test.ts
@@ -7,7 +7,7 @@ describe('api', () => {
       const ok = false;
       const response: Response = {
         url: 'http://localhost:8080/-/packages',
-        ok: ok,
+        ok,
         headers: new Headers(),
         text: async () => responseText,
       } as Response;
--- a/src/utils/api.ts
+++ b/src/utils/api.ts
@@ -6,7 +6,7 @@ import '../../types';
  * @param {object} response
  * @returns {promise}
  */
-export function handleResponseType(response: Response): Promise<[boolean, Blob | string] | void> {
+export function handleResponseType(response: Response): Promise<[boolean, any]> {
   if (response.headers) {
     const contentType = response.headers.get('Content-Type');
     if (contentType && contentType.includes('application/pdf')) {
@@ -26,7 +26,7 @@ export function handleResponseType(response: Response): Promise<[boolean, Blob |
     }
   }
 
-  return Promise.resolve();
+  return Promise.all([response.ok, response.text()]);
 }
 
 class API {
@@ -36,10 +36,11 @@ class API {
     }
 
     const token = storage.getItem('token');
-    if (token && options.headers && typeof options.headers['Authorization'] === 'undefined') {
-      options.headers = Object.assign({}, options.headers, {
-        ['Authorization']: `Bearer ${token}`,
-      });
+    const headers = new Headers(options.headers);
+
+    if (token && headers.has('Authorization') === false) {
+      headers.set('Authorization', `Bearer ${token}`);
+      options.headers = headers;
     }
 
     if (!['http://', 'https://', '//'].some(prefix => url.startsWith(prefix))) {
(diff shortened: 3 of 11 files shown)


juanpicado:
Great move :) 👏 ... let's see how material-ui reacts on this.

src/App/Header/Header.tsx
```diff
@@ -22,9 +22,9 @@ interface Props {
 const Header: React.FC<Props> = ({ withoutSearch }) => {
   const { t } = useTranslation();
   const appContext = useContext(AppContext);
-  const [isInfoDialogOpen, setOpenInfoDialog] = useState();
-  const [showMobileNavBar, setShowMobileNavBar] = useState();
-  const [showLoginModal, setShowLoginModal] = useState(false);
+  const [isInfoDialogOpen, setOpenInfoDialog] = useState<boolean>(false);
```
priscilawebdev, line 25: sometimes when TypeScript can’t infer the type, it makes sense to pass a  generic parameter, but in most of the cases TS can cleverly infer the type for useState, especially when we pass a default value... So I believe that passing a generic parameter here is not really necessary + it's very verbose

src/utils/api.test.ts
```diff
@@ -3,16 +3,18 @@ import api, { handleResponseType } from '../../src/utils/api';
 describe('api', () => {
   describe('handleResponseType', () => {
     test('should handle missing Content-Type', async () => {
+      const responseText = `responseText`;
+      const ok = false;
       const response: Response = {
         url: 'http://localhost:8080/-/packages',
-        ok: false,
+        ok: ok,
```
priscilawebdev, line 10:
```suggestion
        ok,
```

src/utils/api.test.ts
```diff
@@ -3,16 +3,18 @@ import api, { handleResponseType } from '../../src/utils/api';
 describe('api', () => {
   describe('handleResponseType', () => {
     test('should handle missing Content-Type', async () => {
+      const responseText = `responseText`;
+      const ok = false;
```
priscilawebdev, line 7: you could move this line to the object below 😉

src/App/Header/Header.tsx
tmkn: While I agree, in some `useState` calls they were necessary. So in order to not have diverging coding styles I unified the `useState` declarations in places that were affected. Also while they not seem helpful right now, because the code works, TypeScript is now able to error when you try to initialize a state with the wrong type thanks to the generic usage, after all this PR is all about increasing the type safety 😃

src/utils/api.test.ts
tmkn: Yes but I kept them separate because `responseText` and `ok` are part of the assertion, so I declared them together at the top, also so that it's obvious what are the input parameters for the unit test.

src/App/Header/Header.tsx
juanpicado: While Typescript infers values, it does based on how is initialized. So without type anyone could change to `123` and will be hard to notice.

Screen Shot 2020-07-03 at 7 57 25 AM

while the type, perfectly is noticeable and any contributor is aware can only use booleans and not any other value.

Screen Shot 2020-07-03 at 7 57 10 AM

juanpicado: Anyway, I could not find a eslint rule for this, so, let's use common sense. Let's use types for complex objects while for native let's Typescript do the magic.

src/utils/api.test.ts
juanpicado: `ok` is being used on the `response` object. Cannot be below of it.

src/utils/api.ts
```diff
@@ -6,7 +6,7 @@ import '../../types';
  * @param {object} response
  * @returns {promise}
  */
-export function handleResponseType(response: Response): Promise<[boolean, Blob | string] | void> {
+export function handleResponseType(response: Response): Promise<[boolean, any]> {
```
juanpicado, line 9: Just wondering why do we remove types here.

src/utils/api.ts
```diff
@@ -26,7 +26,7 @@ export function handleResponseType(response: Response): Promise<[boolean, Blob |
     }
   }
 
-  return Promise.resolve();
+  return Promise.all([response.ok, response.text()]);
```
juanpicado, line 29: Good catch.

src/App/Header/Header.tsx
juanpicado: And I refer to `useState`. For instance things like `Promise<string>` make totally sense.

src/utils/api.ts
tmkn: Because it also returns json:
https://github.com/verdaccio/ui/blob/f84571251912aae71463c44822c84ca70f8b8c40/src/utils/api.ts#L16
So in this case it returns the tuple `[boolean, any]` since `.json()` returns `any`.

So to be technically correct the return type should be `Promise<[boolean, Blob | string | any]>`, but if you have a union type like this and at any part there is `any`, it will collapse to `any`.

Now you could write it like `Promise<[boolean, Blob | string | any]>` but then 
https://github.com/verdaccio/ui/blob/f84571251912aae71463c44822c84ca70f8b8c40/src/utils/api.ts#L59
errors, since the signature for the `request` method looks like this:
https://github.com/verdaccio/ui/blob/f84571251912aae71463c44822c84ca70f8b8c40/src/utils/api.ts#L33
You would then need a typecast at `resolve(response[1] as any);` which I tried to avoid or go down a rabbit hole of type errors/potential rewrite of the ajax logic.

As I said the ajax functionality is a little bit cumbersome to type as it is. The main cause being that the return type is determined dynamically at runtime (in the `handleResponseType` function), thus there is no natural way for TypeScript to tell the actual return type without explicit casts for the current code. It would help if there were separate ajax functions for json, blob and text, so TypeScript could provide better type safety.
Because right now, the ajax usage is not very type safe:
https://github.com/verdaccio/ui/blob/f84571251912aae71463c44822c84ca70f8b8c40/src/utils/login.ts#L66
The server could return a `string`, in which case there is no code to handle the case that it actually returned a `string` but `json` was expected. 

So in short, without introducing too many type casts/changes to the existing logic, I changed it to `any` to provide a minimum of type safety with the current layout of the code.

src/App/Header/Header.tsx
tmkn: to add there is a subtle difference in the inferred typings for `useState`
used like this: `useState<boolean>();` the type will be `boolean | undefined` while
`useState<boolean>(true);` the type will be just `boolean`

I just like to be more explicit. e.g.
```typescript
const getValue = () => true;
const [useFlag, setUseFlag] = useState(getValue());
```
if `getValue` returns a `string`, then `useState` will not error even though we expect a `boolean`, it will happily infer the `string` type and will instead error further down below, while if `useState` would have made use of the generic to specify the type, it would have error'd.

src/utils/api.ts
juanpicado: Thanks @tmkn :) well explained.

juanpicado:
I'd qualify this as `feat` due the changes might be considerable due the typings in the future.

PR #510, opened 2020-07-03
build: update nvmrc and use latest by default

juanpicado:
I think we might start to use v14 for development by default and forget `12`. All tooling is compatible and way faster.

Remove unnecessary testing on few platforms and versions, this will make builds way faster. We rely on` webpack`, so as long webpack support v14 we are good.

This is specially slow with dependabot, while overload our registry with unnecessary request.


.github/workflows/main.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node_version: [12, 14]
-        os: [ubuntu-latest, windows-latest, macOS-latest]
+        node_version: [14]
+        os: [windows-latest, macOS-latest]
```
sergiohgz, line 13: Why would we want to remove ubuntu from Github CI? 
I think it would be better to have it here and, in the near future, move from CircleCI to GH to remove one of them. What do you think about @juanpicado ?

juanpicado: The idea behind removing it is because, do we need to test all our stack in multi platform? If webpack works or not on Ubuntu well, not our issue, we have automated builds and all already run in unix.

The outcome of the webpack will be exactly the same, so why run test in all platforms? Jest should works in any platform, webpack same, Typescript also, they test already before ship releases.

So, now the `yarn install` is running in each container, does it makes any difference? Nop.

The benefit is not noticeable, but the builds takes for ever to finish. 

I want to advocate everything uses `nvm` to build this project, if someone wants to use Node 10, well, :) have fun.

juanpicado: Perhaps I should update the README to make it clear.

sergiohgz: I agree with your arguments, if a tool (like Webpack or tsc) runs in a system properly or not is not a use case of our CI, but building, testing and lining code in that platform is.

Github actions was very limit to had a great workflow when we create the CI workflow, but now this is not the case. I want to make something in GH like we did in CircleCI, make some stages of the workflow run in only one system (like installing deps) and others in all systems we could to ensure  that using or developing our tool will build in their computers.

If you let me, I can make some PoC to investigate how to improve the workflow this way, caching to make them faster and run some stages only on selected systems to make it faster. Other improve to try is unifying CIs, but that's another story :wink:

juanpicado: Ok go ahead :)

PR #172, opened 2019-10-10
feat: add slim mode

dviryamin:
**Type:** Feature

The following has been addressed in the PR:

*  There is a related issue? yes, verdaccio/verdaccio#1403

**Description:**
Adds welcome page for slim mode instead of long list of the packages.
Configurable using web->mode = slim in config.yaml

Related verdaccio/verdaccio#1516

Resolves: verdaccio/verdaccio#1403

PR #512, opened 2020-07-05
Feat(packagelist): Add list windowing

priscilawebdev:
**Type:** Feat

The following has been addressed in the PR: https://github.com/verdaccio/ui/pull/172

*  There is a related issue? https://github.com/verdaccio/verdaccio/issues/1403
*  Unit or Functional tests are included in the PR: Not yet

**Description:**

After some research, we decided to use the library `react-virtualized`, which despite being large, seems more stable, being maintained by several contributors and used in many different projects and companies.

This PR introduces this library to the `PackageList` component. The list is now being virtualized, providing better performance when rendering a large number of items.

The implementation has some limitations. If the user performs a fast scrolling, some blank white spaces will be shown. This issue seems to be a well-knowen issue and "the only way to ensure that there is no empty space when scrolling is to force scrolling to run on the main/UI thread. " You can read more about it here https://github.com/bvaughn/react-window/issues/43

**Next Steps:**
Evaluate if we can introduce the `<InfiniteLoader/>` component (also from react-virtualized). According to their docs the infinite loader component is: A component that manages just-in-time fetching of data as a user scrolls up or down in a list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "react-hot-loader": "4.12.21",
     "react-i18next": "11.3.1",
     "react-router-dom": "5.1.2",
+    "react-virtualized": "9.21.2",
     "request": "2.88.0",
     "resolve-url-loader": "3.1.1",
     "rimraf": "3.0.0",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@types/react-autosuggest": "9.3.13",
     "@types/react-dom": "16.9.4",
     "@types/react-router-dom": "5.1.5",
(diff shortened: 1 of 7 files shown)


juanpicado:
@priscilawebdev are you planning to have several PR ? .. what's the strategy to follow?

priscilawebdev:
@juanpicado I will only create follow-up PRs if we have decided to introduce an infinite load "pagination" ... on the contrary, that's all :)

juanpicado:
White spaces is ok, we gain in performance. Let's try this for a while and see what happens.

PR #549, opened 2020-09-13
 ref(icons): Update icons

priscilawebdev:
Converted Svg files into JSX Files +  Added the country names that we use in the translation files


juanpicado:
Could you share why do you want to *Converted Svg files into JSX Files *?

priscilawebdev:
@juanpicado I would like to have them as a component, having control over it

DanielRuf:
I'm triggering all builds again. There seems to be some ntwork related issue with GHA. Please hang on.

DanielRuf:
We should disable prettier for these files (generally I hate prettier).

Bildschirmfoto 2020-09-16 um 00 33 38

I guess after these errors are resolved the builds should be green.

DanielRuf (approved):
Awesome =)

LGTM

PR #550, opened 2020-09-16
chore: add yarn timeout setting to ci

juanpicado:
increase yarn timeout for ci


DanielRuf:
> is failing to fetch deps randomly and potentially block a release.

Do you mean the network related issues with yarn or something else?

priscilawebdev (dismissed):

DanielRuf:
> Do you mean the network related issues with yarn or something else?

Because we can solve that with a yarn option. See the commit by me at https://github.com/verdaccio/ui/pull/549/commits

https://github.com/verdaccio/ui/pull/549/commits/b1b51dc265c78162ee4e5096c9c8f20afa75faa2

I'm not sure if we should stop testing on Windows. Because then we can not test if there are other unrelated issues with Windows and verdaccio. I would not drop Windows from the CI setup but increase the timeout like I did in the linked commit @juanpicado. There is always a solution for the much slower Windows hosts.

juanpicado:
I'll try that @DanielRuf thanks for that !

verdacciobot (approved):

Issue #93, opened 2019-07-08
Verify wether identity-obj-proxy in Jest works


juanpicado:
We should not mock this due to we uses `identity-obj-proxy` to do it out of the box, but If I remove the mock the test fails, thus, the configuration seems does not work fine. I'll address this in a new ticket.
https://github.com/keyz/identity-obj-proxy
See example here.
https://github.com/verdaccio/ui/blob/4.x-master/jest.config.js#L16

_Originally posted by @juanpicado in https://github.com/verdaccio/ui/pull/92/review_comment/create_

sergiohgz:
In the example referenced at https://github.com/keyz/identity-obj-proxy, the images are mocked with a stub file config. I don't see any site where they use identity-obj-proxy for this purpose, so we would make the workaround. What do you think? @juanpicado @ayusharma

juanpicado:
Yeah, I'm confused about the current usage. https://jestjs.io/docs/en/webpack#mocking-css-modules Even they only reference webpack or CSS modules and that's is what @griffithtp is removing here https://github.com/verdaccio/ui/pull/85 , so perhaps we do not need that anymore and @ayusharma is right with this approach.

sergiohgz:
We can use it to mock svg files.
Currently, the svg transformer in https://github.com/verdaccio/ui/blob/4.x-master/jest.config.js#L17 is not working because it doesn't exist. But if we use `identity-obj-proxy` (we can join s?css and svg definition in line 14) we don't need to mock svg files anymore like we did in https://github.com/verdaccio/ui/blob/4.x-master/src/components/Engines/Engines.test.tsx#L5

priscilawebdev:
closing this for now

PR #572, opened 2020-10-09
build: migrate from yarn v1 to  yarn berry (v2)

### LINKED ISSUE #558: Migrate to Yarn v2
- [ ] I'd be willing to implement the feature

**Is your feature request related to a problem?**

Verdaccio UI uses Yarn v1. Since this is not a monorepo and we want to learn more about Yarn Berry, we want to run Yarn v2 on this repo from now on. 

**Describe the solution you'd like**

Full migration, with zip files and PnP enabled.

juanpicado:
**Type:** build

**Description:**

For an easy upgrade, we are still using `nodeLinker: node-modules` which is almost the same as `yarn v1`. The idea is move to PnP as soon are issues are fixed. We would need to update webpack and other small issues I found.

> :warning: `security/snyk` fails, this is a known issue, cannot do anything about it but will be fixed eventually. 

- [x] Update CI
- [x]  Migrate to yarn
- [x] Fix test configuration
- [x] Fix lint
- [x] Check build works fine
- [x] Update contribution guidelines
- [x] fix `yarn dev` (it gets hang)
- [x] fix e2e (few test are failing)

 Resolves #558


.yarn/sdks/typescript/lib/tsserver.js
```diff
@@ -0,0 +1,77 @@
...
+
+  return Object.assign(Session.prototype, {
+    onMessage(/** @type {string} */ message) {
+      const parsedMessage = JSON.parse(message)
+
+      if (
+        parsedMessage != null &&
+        typeof parsedMessage === 'object' &&
+        parsedMessage.arguments &&
+        parsedMessage.arguments.hostInfo === 'vscode'
```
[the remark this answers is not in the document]
priscilawebdev: or `isVSCode = parsedMessage?.arguments?.hostInfo === 'vscode'`

.yarn/sdks/typescript/lib/tsserver.js
```diff
@@ -0,0 +1,77 @@
...
+    // We add the `zip:` prefix to both `.zip/` paths and virtual paths
+    if (isAbsolute(str) && !str.match(/^\^zip:/) && (str.match(/\.zip\//) || str.match(/\$\$virtual\//))) {
+      // Absolute VSCode `Uri.fsPath`s need to start with a slash.
+      // VSCode only adds it automatically for supported schemes,
+      // so we have to do it manually for the `zip` scheme.
+      // The path needs to start with a caret otherwise VSCode doesn't handle the protocol
+      // https://github.com/microsoft/vscode/issues/105014#issuecomment-686760910
+      return `${isVSCode ? '^' : ''}zip:${str.replace(/^\/?/, `/`)}`;
+    } else {
+      return str;
```
priscilawebdev, line 58:
```suggestion
}
      return str;
```

.yarn/sdks/typescript/lib/typescript.js
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
+
+const {existsSync} = require(`fs`);
+const {createRequire, createRequireFromPath} = require(`module`);
+const {resolve, dirname} = require(`path`);
+
+const relPnpApiPath = "../../../../.pnp.js";
+
+const absPnpApiPath = resolve(__dirname, relPnpApiPath);
+const absRequire = (createRequire || createRequireFromPath)(absPnpApiPath);
```
priscilawebdev, line 10: I noticed that we are repeating this in a couple of places... maybe we should make it reusable?

babel.config.js
```diff
@@ -0,0 +1,28 @@
+const BROWSER_TARGETS = {
+  browsers: ['last 5 versions', 'FireFox >= 44', 'Safari >= 7', 'Explorer 11', 'last 4 Edge versions'],
```
priscilawebdev, line 2: at the beginning of  August 17, 2021, Microsoft will no longer support IE 11..should we wait?

jest/identity.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable no-var, comma-dangle, no-unused-vars, global-require */
+let Reflect;
+let idObj;
+
+function checkIsNodeV6OrAbove() {
+  if (typeof process === 'undefined') {
```
priscilawebdev, line 6:
```suggestion
  if (process === undefined) {
```

jest/identity.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable no-var, comma-dangle, no-unused-vars, global-require */
+let Reflect;
```
priscilawebdev, line 2: Maybe we should add a description explaining what this file does?

package.json
```diff
@@ -13,18 +13,42 @@
...
+    "@babel/plugin-proposal-decorators": "7.10.5",
+    "@babel/plugin-proposal-export-namespace-from": "7.10.4",
+    "@babel/plugin-proposal-function-sent": "7.10.4",
+    "@babel/plugin-proposal-json-strings": "7.10.4",
     "@babel/plugin-proposal-nullish-coalescing-operator": "7.10.4",
+    "@babel/plugin-proposal-numeric-separator": "7.10.4",
+    "@babel/plugin-proposal-object-rest-spread": "7.11.0",
     "@babel/plugin-proposal-optional-chaining": "7.11.0",
-    "@babel/polyfill": "7.11.5",
+    "@babel/plugin-proposal-throw-expressions": "7.10.4",
```
priscilawebdev, line 28: Please are these new dependencies necessary for this PR?

package.json
```diff
@@ -166,32 +199,26 @@
...
-    "lint:css": "stylelint \"src/**/styles.ts\"",
-    "lint:lockfile": "lockfile-lint --path yarn.lock --type yarn --validate-https --allowed-hosts verdaccio npm yarn",
+    "lint": "yarn run lint:js && yarn run lint:css",
+    "lint:js": "yarn run type-check && eslint . --ext .js,.ts,.tsx",
+    "lint:html": "yarn htmllinter src/template/index.html",
+    "lint:css": "yarn stylelint \"src/**/styles.ts\"",
     "coverage:publish": "codecov",
-    "pre:webpack": "rimraf static/*",
-    "dev:web": "cross-env BABEL_ENV=ui babel-node tools/dev.server.js",
+    "dev:web": "babel-node tools/dev.server.js",
```
priscilawebdev, line 216: it looks much more clear :)

package.json
```diff
@@ -219,5 +246,13 @@
     "type": "opencollective",
     "url": "https://opencollective.com/verdaccio",
     "logo": "https://opencollective.com/verdaccio/logo.txt"
+  },
+  "dependenciesMeta": {
```
priscilawebdev, line 250: Please why this change is necessary?

babel.config.js
priscilawebdev: Please, does it work? shouldn't it be better to add a browser slug and version?

.yarn/sdks/typescript/lib/tsserver.js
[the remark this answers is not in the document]
juanpicado: All inside of `.sdks` is generated code.  https://yarnpkg.com/getting-started/editor-sdks

package.json
juanpicado: Yes, I removed `@verdaccio/babel-preset`, so the babel dependencies need to be on the root from now on. Yarn 2 is strict on this matter.

package.json
juanpicado: The reason of those was because `@verdaccio/babel-preset`, so not need it anymore.

package.json
juanpicado: Only if we use `pnp`, so in this case are not need it, but this command https://yarnpkg.com/cli/unplug#gatsby-focus-wrapper actually create the new `dependenciesMeta`, is a new field introduced by yarn. 

I'll remove it for now.

.yarn/sdks/typescript/lib/typescript.js
juanpicado: Ignore all inside of `sdks`. I'll add it to `.gitignore`, I think those should be downloaded via vcode.

babel.config.js
juanpicado: It does work, Babel does it out of the box including few polyfills, like `fetch`. Remove that support won't change much for us to be honest, thus positive impact is almost 0. I'd prefer keep it as long material-ui does it https://material-ui.com/getting-started/supported-platforms/

jest/identity.js
juanpicado: I removed one dependency `identity-obj-proxy` and copy the code there, yarn2 did not like it and was breaking all test. https://github.com/verdaccio/ui/issues/93 meant to verify whether still need this, so I might left a comment on the ticket so anyone can match this change.

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
+    "arcanis.vscode-zipfs",
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
```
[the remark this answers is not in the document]
juanpicado: This will install the `sdks` extensions, so we don't need that code commited.

.yarn/sdks/typescript/lib/typescript.js
merceyz: If you're using PnP it should be commited, if not you can just remove it :)

.vscode/settings.json
```diff
@@ -1,15 +1,23 @@
...
+  },
+  "eslint.alwaysShowStatus": true,
+  "search.exclude": {
+    "**/.yarn": true,
+    "**/.pnp.*": true
+  },
+  "eslint.nodePath": ".yarn/sdks",
+  "prettier.prettierPath": ".yarn/sdks/prettier/index.js",
+  "typescript.enablePromptUseWorkspaceTsdk": true,
+  "stylelint.stylelintPath": ".yarn/sdks/stylelint/lib/index.js"
```
merceyz, line 22: Since you're not using PnP and you removed the SDK you need to remove these otherwise the language server will crash

package.json
```diff
@@ -166,32 +199,26 @@
...
-    "test:clean": "npx jest --clearCache",
+    "test:clean": "jest --clearCache",
     "test:acceptance": "codeceptjs run --steps",
-    "test:acceptance:server": "concurrently --kill-others \"npm run verdaccio:server\" \"npm run test:acceptance\"",
-    "test:e2e": "cross-env BABEL_ENV=test jest --config ./test/jest.config.e2e.js",
-    "test": "cross-env NODE_ENV=test BABEL_ENV=test TZ=UTC jest --config ./jest/jest.config.js --maxWorkers 2 --passWithNoTests",
-    "test:update-snapshot": "npm run test -- -u",
+    "test:acceptance:server": "concurrently --kill-others \"yarn run verdaccio:server\" \"yarn run test:acceptance\"",
+    "test:e2e": "cross-env jest --config ./test/jest.config.e2e.js",
+    "test": "cross-env NODE_ENV=test TZ=UTC jest --config ./jest/jest.config.js --maxWorkers 2 --passWithNoTests",
```
merceyz, line 208: You don't need `cross-env` with V2, it supports this out of the box

merceyz (commented):
Seems the `.yarn/install-state.gz` file snuck past your gitignore https://github.com/verdaccio/ui/pull/572/files#diff-05a728841a7e7287c563b47f8a43e30b

package.json
merceyz: You can still use the preset but it needs to use `require.resolve` for the plugins/presets it adds

.yarn/sdks/typescript/lib/typescript.js
juanpicado: Thanks really appreciated feedback. How this deal with conflicts? I guess eventually the code might be updated.

.vscode/settings.json
juanpicado: good catch ! thanks

.yarn/sdks/typescript/lib/tsserver.js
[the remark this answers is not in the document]
priscilawebdev: oh yes :)

priscilawebdev (approved):
I'd just keep using `@verdaccio/babel-preset ` and use `require.resolve `as @merceyz suggested because I think the dependencies are clearer this way ...but other than that it looks good to me...Good Job 👏

.yarn/sdks/typescript/lib/typescript.js
merceyz: When you update Yarn you'd also run `yarn dlx @yarnpkg/pnpify --sdk vscode` to update it, but unless something changes from vscode you most likely wont have to.

.github/workflows/canary.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           node-version: '12.x'
           registry-url: 'https://registry.verdaccio.org'
       - name: Install
-        run: yarn install --frozen-lockfile --network-timeout 100000
+        run: yarn install
```
merceyz, line 20: `--frozen-lockfile` is now `--immutable` https://yarnpkg.com/cli/install#options

juanpicado: thanks ! I did not see this one :)

package.json
juanpicado: Thanks, on this specific case, I will use explicit dependencies, much clear for contributors what we are using and easier to update or replace.

juanpicado:
Thanks guys for the review :guitar:

PR #587, opened 2020-12-09
Update zh-TW.json

FatDoge:
**Type:**

*  Micro docs fix.

**Description:**
* As a province, taiwan is a part of China.


juanpicado:
@FatDoge thanks for your contribution, but I don't get in politics. Let's keep things as it is. I'll respectfully close this.

PR #595, opened 2021-01-12
Remove capitalisation from greeting label

imjacobclark:
**Type:**

Removing capitalisation from greeting label. When Verdaccio is welcoming a user based on their username, it is odd to capitalise the first letter, it is probably better to make zero assumptions about the formatting of the name rather than try and format it ourselves on the UI and simply accept whatever the auth source provides as a source of truth.

In the attached screenshot, you can see the strangeness I mention, "Imjacobclark" opposed to "imjacobclark".

Screenshot 2021-01-12 at 15 28 14

### DIFF
--- a/src/App/Header/HeaderGreetings.tsx
+++ b/src/App/Header/HeaderGreetings.tsx
@@ -14,7 +14,7 @@ const HeaderGreetings: React.FC<Props> = ({ username }) => {
   return (
     <>
       <Greetings>{t('header.greetings')}</Greetings>
-      <Label capitalize={true} data-testid="greetings-label" text={username} weight="bold" />
+      <Label data-testid="greetings-label" text={username} weight="bold" />
     </>
   );
 };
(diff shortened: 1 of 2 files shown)


chrisgrounds:
I actually like this feature. Can we have the option to choose whether we want capitalisation or not? Could you code that in as well @imjacobclark ?

imjacobclark:
> I actually like this feature. Can we have the option to choose whether we want capitalisation or not? Could you code that in as well @imjacobclark ?

I think first would be to resolve for the simplest case and then potentially raise a ticket to customise in future.

priscilawebdev (approved):
I agree with you @imjacobclark Thanks for the PR 😉

imjacobclark:
> I agree with you @imjacobclark Thanks for the PR 😉

Thanks @priscilawebdev! What's required to get this change merged? 

Thanks!

PR #606, opened 2021-03-04
feat!: reduce bundle size webpack improvement

juanpicado:
**Type:** build
**Description:**

Many issues reported using `prefix_url`  https://verdaccio.org/docs/en/configuration#url-prefix are related the big amount of assets to be loaded. This remove the `url()` from CSS files (or CSS at all) so we don't need to worry about that anymore. This is the first required step to solve that issue. 

Also uses new webpack 5 asset loader which comes built in. 

Reduce verdaccio size in few hundred kb. 

```
static
├── 238.35e8de97062729a5ca05.js
├── 238.35e8de97062729a5ca05.js.LICENSE.txt
├── 818.35e8de97062729a5ca05.js
├── 818.35e8de97062729a5ca05.js.LICENSE.txt
├── Dependencies.35e8de97062729a5ca05.js
├── Dependencies.35e8de97062729a5ca05.js.LICENSE.txt
├── Dist.35e8de97062729a5ca05.js
├── Dist.35e8de97062729a5ca05.js.LICENSE.txt
├── Engines.35e8de97062729a5ca05.js
├── Engines.35e8de97062729a5ca05.js.LICENSE.txt
├── favicon.ico
├── Home.35e8de97062729a5ca05.js
├── Home.35e8de97062729a5ca05.js.LICENSE.txt
├── index.html
├── Install.35e8de97062729a5ca05.js
├── Install.35e8de97062729a5ca05.js.LICENSE.txt
├── main.35e8de97062729a5ca05.js
├── main.35e8de97062729a5ca05.js.LICENSE.txt
├── NotFound.35e8de97062729a5ca05.js
├── NotFound.35e8de97062729a5ca05.js.LICENSE.txt
├── Provider.35e8de97062729a5ca05.js
├── Provider.35e8de97062729a5ca05.js.LICENSE.txt
├── Repository.35e8de97062729a5ca05.js
├── Repository.35e8de97062729a5ca05.js.LICENSE.txt
├── runtime.35e8de97062729a5ca05.js
├── runtime.35e8de97062729a5ca05.js.LICENSE.txt
├── UpLinks.35e8de97062729a5ca05.js
├── UpLinks.35e8de97062729a5ca05.js.LICENSE.txt
├── vendors.35e8de97062729a5ca05.js
├── vendors.35e8de97062729a5ca05.js.LICENSE.txt
├── Version.35e8de97062729a5ca05.js
├── Version.35e8de97062729a5ca05.js.LICENSE.txt
├── Versions.35e8de97062729a5ca05.js
└── Versions.35e8de97062729a5ca05.js.LICENSE.txt
```

#### Related tickets

https://github.com/verdaccio/verdaccio/issues/1523
https://github.com/verdaccio/verdaccio/issues/1297
https://github.com/verdaccio/verdaccio/issues/1593
https://github.com/verdaccio/verdaccio/discussions/1539
https://github.com/verdaccio/website/issues/264
https://github.com/verdaccio/verdaccio/issues/1565
https://github.com/verdaccio/verdaccio/issues/1251
https://github.com/verdaccio/verdaccio/issues/2029
https://github.com/verdaccio/docker-examples/issues/29

### DIFF
--- a/src/design-tokens/theme.ts
+++ b/src/design-tokens/theme.ts
@@ -84,7 +84,13 @@ export const getTheme = (themeMode: ThemeMode) => {
   const palette = themeModes[themeMode];
   return createMuiTheme({
     typography: {
-      fontFamily: 'inherit',
+      fontFamily: [
+        '-apple-system',
+        'BlinkMacSystemFont',
+        '"Helvetica Neue"',
+        'Arial',
+        'sans-serif',
+      ].join(','),
     },
     palette: {
       type: themeMode,
--- a/src/design-tokens/theme.ts
+++ b/src/design-tokens/theme.ts
@@ -84,13 +84,7 @@ export const getTheme = (themeMode: ThemeMode) => {
   const palette = themeModes[themeMode];
   return createMuiTheme({
     typography: {
-      fontFamily: [
-        '-apple-system',
-        'BlinkMacSystemFont',
-        '"Helvetica Neue"',
-        'Arial',
-        'sans-serif',
-      ].join(','),
+      fontFamily: ['-apple-system', 'BlinkMacSystemFont', '"Helvetica Neue"', 'Arial', 'sans-serif'].join(','),
     },
     palette: {
       type: themeMode,
--- a/package.json
+++ b/package.json
@@ -179,18 +179,6 @@
     {
       "path": "./static/main.*.js",
       "maxSize": "400 kB"
-    },
-    {
-      "path": "./static/*.svg",
-      "maxSize": "1.3 kB"
-    },
-    {
-      "path": "./static/*.png",
-      "maxSize": "15 kB"
-    },
-    {
-      "path": "./static/fonts/*.{woff,woff2}",
-      "maxSize": "20 kB"
     }
   ],
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -150,7 +150,6 @@
     "stylelint-webpack-plugin": "^2.1.1",
     "supertest": "4.0.2",
     "terser-webpack-plugin": "^5.0.3",
-    "typeface-roboto": "0.0.75",
     "typescript": "3.9.6",
     "url-loader": "^4.1.1",
     "validator": "13.1.1",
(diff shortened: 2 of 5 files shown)


src/design-tokens/theme.ts
```diff
@@ -84,7 +84,7 @@ export const getTheme = (themeMode: ThemeMode) => {
   const palette = themeModes[themeMode];
   return createMuiTheme({
     typography: {
-      fontFamily: 'inherit',
+      fontFamily: ['-apple-system', 'BlinkMacSystemFont', '"Helvetica Neue"', 'Arial', 'sans-serif'].join(','),
```
priscilawebdev, line 87: do we really want to change our font-family? I kinda liked Roboto 😁

juanpicado: Yeah, I wana keep it but it brings a lot of headaches for people using proxies. The main issue is the `relative path` is generated in build time by webpack eg: `url(/static/xxx.xx)` this is a big issue if a reverse proxy has a relative path instead root path. I've tried to fix it last year but did not work out, I have a piple of duplicated issues that report that. So my solution is get rid of the assets, we use Emotion and the `<styles>` are being generated with JS, images and svg are base64 by default, so the Fonts were the next barrier, actually looks the same, I think we can precint of them. 

There is another thing I need to do in a major release, is a single JS server with the UI emebebed, to achieve this also you cannot have fonts, so, they will be gone somehow eventually anyway. 

My next step, is restore what I broke on `v4.1.2` on the next minor and try to close all those tickets.

priscilawebdev: Oh, I understand. Yes, if removing the fonts helps to resolve the issue, then it is better to remove them, and we can maybe try to find another solution soon. Thanks for the great explanation!

priscilawebdev (approved):

src/design-tokens/theme.ts
juanpicado: I'll keep you posted :)

PR #607, opened 2021-03-11
build: visual snapshots only on pull request

juanpicado:
Small improvement, it will run visual snapshots only on pull request.

We should not run external services on master that might force a release failure, furthermore we don't work on master only via pull request, thus so it won't have much impact.


priscilawebdev (approved):
thank you for this :)

PR #608, opened 2021-03-13
feat: api provider and config provider

juanpicado:
**Description:**

```
<AppConfigurationContext>
  <APIProvider>
     ---> The test of the App <--- <App />
  </APIProvider>
</AppConfigurationContext>
```

Introduce two new providers, 

- Add base html tag thus webpack is able to handle properly the routing
- the `AppConfigurationContext` which holds the variables which are embedded at the html, then using
` const { configOptions } = useConfig();` via context can easily access each property.
- the `APIProvider` which provides all API methods, one single place to rule all request instead have different references to API in different files eg: `const { callDetailPage, callReadme } = useAPI();`.
- Unique global `window.__VERDACCIO_BASENAME_UI_OPTIONS` to rule all options, remove old global variables
- Improve validations (specially for theme primary color, material ui crash otherwise)

This is part of https://github.com/verdaccio/verdaccio/pull/2122


.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": [
     "@verdaccio",
+    "prettier",
```
juanpicado, line 4: this was missing

package.json
```diff
@@ -212,11 +212,9 @@
   },
   "lint-staged": {
     "*.{js,tsx,ts}": [
-      "eslint . --ext .js,.ts,.tsx",
-      "prettier --write"
+      "eslint . --ext .js,.ts,.tsx"
     ],
     "*": [
-      "detect-secrets-launcher --baseline .secrets-baseline --word-list",
```
juanpicado, line 219: I disable this while I find a way to ignore some stuff

src/providers/API/APIProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+      body: JSON.stringify({ username, password }),
+      headers: {
+        Accept: HEADERS.JSON,
+        'Content-Type': HEADERS.JSON,
+      },
+    });
+    const result: LoginBody = {
+      username: response.username,
+      token: response.token,
+    };
```
priscilawebdev, line 82: if the response has the same types of result, wouldn't it be better if just return  `await API.request(buildURL('login'), 'POST', {` ?

src/providers/API/APIProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+    return result;
+  };
+
+  const getResource = async (link: string): Promise<Blob> => {
+    const fileStream: Blob = await API.request(link, 'GET', {
```
priscilawebdev, line 87: maybe we could just return the promise and if needed update the function name to be more descriptive?

src/providers/API/APIProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+  );
+
+  return <AppAPIContext.Provider value={value}>{children}</AppAPIContext.Provider>;
+};
+
+export default APIProvider;
+
+const useAPI = () => useContext(AppAPIContext);
+
+export { useAPI };
```
priscilawebdev, line 117:
```suggestion
export const useAPI = () => useContext(AppAPIContext);
```

src/providers/API/storage.ts
```diff
@@ -0,0 +1,12 @@
+import memoryStorage from 'localstorage-memory';
+
+let storage: Storage;
```
priscilawebdev, line 3: should we initiate `storage` with the `localStorage` or with `undefined`?

src/providers/config/AppConfigurationProvider.tsx
```diff
@@ -0,0 +1,68 @@
+import isEmpty from 'lodash/isEmpty';
+import isNil from 'lodash/isNil';
+import React, { createContext, FunctionComponent, useContext, useMemo, useState } from 'react';
+
+export type VerdaccioOptions = {
+  url_prefix: string;
+  uri?: string;
```
priscilawebdev, line 7: can we please move the optional props to the bottom of the object? 🙏

src/providers/config/AppConfigurationProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  scope: string;
+  title: string;
+  logo?: string;
+  primaryColor: string;
+  darkMode: boolean;
+};
+
+type ConfigProviderProps = {
+  configOptions: VerdaccioOptions;
+  setConfigOptions: any;
```
priscilawebdev, line 22: is it possible to set the correct type here?

src/providers/config/AppConfigurationProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+};
+
+const defaultValues: ConfigProviderProps = {
+  configOptions: {
+    primaryColor: '#4b5e40',
```
priscilawebdev, line 27: we use the same color hex in several files. Should we maybe create a const and reuse it? then it's easy to update it in the future

src/providers/config/AppConfigurationProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  if (isNil(uiConfiguration.primaryColor) || isEmpty(uiConfiguration.primaryColor)) {
+    uiConfiguration.primaryColor = '#4b5e40';
+  }
+
+  return uiConfiguration;
+}
+
+const AppConfigurationContext = createContext<ConfigProviderProps>(defaultValues);
+
+const AppConfigurationProvider: FunctionComponent = ({ children }) => {
```
priscilawebdev, line 49: I was wondering if we should keep using `FunctionComponent`.. just thinking :)

src/providers/config/AppConfigurationProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  // @ts-ignore
+  return <AppConfigurationContext.Provider value={value}>{children}</AppConfigurationContext.Provider>;
+};
+
+export default AppConfigurationProvider;
+
+const useConfig = () => useContext(AppConfigurationContext);
+
+export { useConfig };
```
priscilawebdev, line 68:
```suggestion
export const useConfig = () => useContext(AppConfigurationContext);
```

priscilawebdev:
I left some suggestions, but really great job 🚀

src/design-tokens/theme.ts
```diff
@@ -40,6 +40,14 @@ const themeModes = {
   },
 };
 
+function applyPrimaryColor(mode: ThemeMode, primaryColor: string): any {
+  if (mode === 'light') {
+    themeModes['light'].primary = primaryColor;
```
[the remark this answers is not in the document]
juanpicado: Well I just followed what was established, I don't know whether a dark mode can have something different that dark.

src/providers/config/AppConfigurationProvider.tsx
juanpicado: Yes, otherwise children yells `Binding element 'children' implicitly has an 'any' type.ts(7031)`

src/App/Header/LoginDialog/LoginDialog.test.tsx
```diff
@@ -79,25 +81,31 @@ describe('<LoginDialog /> component', () => {
...
     // TODO: the input's value is not being updated in the DOM
     const passwordInput = getByPlaceholderText('Your strong password');
+
     fireEvent.focus(passwordInput);
     fireEvent.change(passwordInput, { target: { value: '1234' } });
 
     // TODO: submitting form does not work
-    const signInButton = getByText('Sign in');
-    fireEvent.click(signInButton);
+    const signInButton = getByTestId('login-dialog-form-login-button');
```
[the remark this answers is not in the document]
juanpicado: There are two matches with Sign In, actually use translations for test might be no the best idea, I'd rather use the original keys which are unique. Anyway using the test id I wanna be sure I'm hitting the button and not something else.

src/providers/API/storage.ts
juanpicado: storage cannot be undefined, the code was just moved. I assume what does try to use local storage but if is not there switch to the memory, but cannot be undefined.

src/App/Header/LoginDialog/LoginDialog.tsx
```diff
@@ -20,6 +23,35 @@ interface Props {
 const LoginDialog: React.FC<Props> = ({ onClose, open = false }) => {
   const { t } = useTranslation();
   const appContext = useContext(AppContext);
+  const { doLogin } = useAPI();
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
juanpicado: Mmm I don't think so, you might end up in a memory leak, no updated methods or endless API calls,
https://github.com/facebook/react/tree/master/packages/eslint-plugin-react-hooks#advanced-configuration

src/design-tokens/theme.ts
```diff
@@ -1,5 +1,7 @@
 import createMuiTheme from '@material-ui/core/styles/createMuiTheme';
 
+import { PRIMARY_COLOR } from 'verdaccio-ui/utils/colors';
```
priscilawebdev, line 3: thanks for that 👏

src/design-tokens/theme.ts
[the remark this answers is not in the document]
priscilawebdev: Hm, I think if we allow the user to customize the primary color of the light theme, maybe we should allow customization of the primary color for the dark theme too, but let's think about it and maybe add this change in a follow-up PR 😉

src/providers/API/storage.ts
priscilawebdev: in this case, should we initiate it with `memoryStorage`? I understand what you mean, but as `let storage: Storage;` is just declared, it's already `undefined` in the beginning

src/providers/config/AppConfigurationProvider.tsx
priscilawebdev: sure thing, but maybe we could write regular functions and add the type children to the Interfaces 🤷‍♀️ it's just an idea.

priscilawebdev (approved):
lgtm Well done 👏

PR #609, opened 2021-03-19
feat: replace base path

juanpicado:
Again regarding https://github.com/verdaccio/verdaccio/pull/2122

By using `ToReplaceByVerdaccio` also replaces `ToReplaceByVerdaccioFavico`, so  must be renamed.

```
 basePath: 'ToReplaceByVerdaccio',
 faviIcoPath: `ToReplaceByVerdaccioFavico`,
```

This will break all E2E but there is no other way, this is one of the reason why I want to move UI to core, I cannot do breaking changes. I need to merge this even if break the test, there is no other way.


juanpicado:
I would need to do bigger changes, I will reopen a new one later on.

PR #612, opened 2021-04-07
fix: inject css files to the dom as style

juanpicado:
Fix issue on Verdaccio 5 , the markdown overflow the detail, the reason was that webpack was not including the css files with the style-loader. 

Here you can see the broken layout.
https://registry.verdaccio.org/-/web/detail/verdaccio


juanpicado:
I merge, need to release this, there is a bug need to fix before release.

